PR #14, opened 2024-02-14
Refactoring: Rename folder `t` to `topic`

amotl:
## About
We received a request to dissolve the shorthand `t` folder.

## Thoughts
The original rationale was to add a grouping level without adding too many characters to the URL, in the same spirit like Discourse does it, e.g. https://community.crate.io/t/deploying-cratedb-to-the-cloud-via-terraform/849.

However, we recognize it might not have been the right decision for a documentation system / repository, so this patch changes that again, using the long version `topic` instead of the shorthand version `t`.


seut (approved):
👍 thx!

hammerhead (approved):
I was just seeing that folder earlier and couldn't make sense of its meaning. Better to have it explicit 👍.

PR #15, opened 2024-02-20
Refactoring: Rename `handbook` to `admin` again

amotl:
## Why?
The label fits better to the content, and it actually has been the label used previously.

## Preview
https://cratedb-guide--15.org.readthedocs.build/
https://cratedb-guide--15.org.readthedocs.build/admin/


docs/admin/index.rst
```diff
@@ -1,10 +1,9 @@
-.. _handbook:
-.. _handbooks:
-.. _user-handbooks:
```
seut, line 3: just double checking: We do not reference this anywhere under these names already, right?

seut (approved):
Lgtm otherwise.

docs/admin/index.rst
amotl: I don't think so, because I've just introduced that recently. Otherwise, referenced projects will tell us on CI.

PR #32, opened 2024-02-21
Style: Make tile gallery on "install" page responsive

amotl:
## Problem
The "install" page did not display well on smaller screens, like mobile phones.

## Solution
Use responsive features of [Bootstrap](https://en.wikipedia.org/wiki/Bootstrap_(front-end_framework)) per [sphinx{design}](https://sphinx-design.readthedocs.io/).

## Preview
Before: https://cratedb.com/docs/guide/install/index.html
After: https://cratedb-guide--32.org.readthedocs.build/install/index.html

image image

/cc @karynzv, @seut, @matriv, @hlcianfagna, @hammerhead, @proddata, @ckurze

### DIFF
--- a/docs/install/index.rst
+++ b/docs/install/index.rst
@@ -38,7 +38,7 @@ Please read this section carefully.
 .. the following sections explicitly
 
 
-.. grid:: 4
+.. grid:: 2 2 2 4
     :padding: 0
     :class-container: installation-grid
 


docs/install/index.rst
```diff
@@ -38,7 +38,7 @@ Please read this section carefully.
 .. the following sections explicitly
 
 
-.. grid:: 4
+.. grid:: 2 2 2 4
```
amotl, line 41: > Grids are based on a 12 column system, which can adapt to the size of the viewing screen.
>
> A grid directive can be set with the number of default columns (1 to 12); either a single number for all screen sizes, or four numbers for extra-small (<576px), small (768px), medium (992px) and large screens (>1200px), then child grid-item directives should be set for each item.

-- https://sphinx-design.readthedocs.io/en/latest/grids.html

matkuliak (approved):
Much better

PR #37, opened 2024-02-22
Theme: Update to crate-docs-theme==0.30.0.dev5

amotl:
## About

Consume a new pre-release version of crate-docs-theme, in order to probe the appearance on cratedb-guide.


amotl:
GH-12 is already there.

PR #62, opened 2024-03-07
Improve grid display on mobile

msbt:

The grid isn't responsive, following [this guide](https://mystmd.org/guide/dropdowns-cards-and-tabs#grids), this should make it 1 col on extra-small devices (<576px), 2 col on small (768px) and 3 col on medium (992px) and large screens (>1200px) - not the first element though, this will stay 2 col.


amotl (approved):
Thanks for fixing this up!

PR #69, opened 2024-04-10
Installation: Update section about installing RPM packages

amotl:
## About

Registering with the RPM package repository using the `crate-release-7.0-1.x86_64.rpm` packages and friends is being deprecated. It is advised to configure the RPM repository manually instead.

## Preview

https://cratedb-guide--69.org.readthedocs.build/install/redhat.html

## Caveat

I've only edited the documentation, **without doing any kind of validation yet.** Maybe you can spare a few minutes? Thanks for your support!

/cc @hlcianfagna, @WalBeh, @BaurzhanSakhariev, @matriv

### DIFF
--- a/docs/install/redhat.rst
+++ b/docs/install/redhat.rst
@@ -9,7 +9,7 @@
 CrateDB on Red Hat, SUSE, and Derivates
 #######################################
 
-Install CrateDB RPM_ packages using the YUM_ package manager.
+Install CrateDB RPM_ packages using the DNF_, YUM_, or ZYpp_ package managers.
 
 This installation method is suitable for RedHat Enterprise Linux (RHEL) and compatible
 systems like Fedora, CentOS, Rocky Linux, AlmaLinux, AWS Linux, Oracle Linux, or
@@ -19,30 +19,37 @@ Scientific Linux. Installation also works on openSUSE and SUSE Linux Enterprise 
 Configure package repository
 ============================
 
-You will need to configure your system to register with and trust packages
-from the CrateDB package repository::
+To register with the CrateDB package repository, create a file called ``cratedb.repo``
+in the ``/etc/yum.repos.d/`` directory for RedHat based distributions, or in the
+``/etc/zypp/repos.d/`` directory for OpenSuSE based distributions, containing::
 
-    # Install prerequisites.
-    yum install sudo
+    [cratedb-ce-stable]
+    name=CrateDB RPM package repository - $basearch - Stable
+    baseurl=https://cdn.crate.io/downloads/yum/7/$basearch
+    enabled=0
+    gpgcheck=1
+    gpgkey=https://cdn.crate.io/downloads/yum/RPM-GPG-KEY-crate
+    autorefresh=1
+    type=rpm-md
 
-    # Import the public GPG key for verifying the package signatures.
-    sudo rpm --import https://cdn.crate.io/downloads/yum/RPM-GPG-KEY-crate
-
-    # Register with the CrateDB package repository.
-    sudo rpm -Uvh https://cdn.crate.io/downloads/yum/7/x86_64/crate-release-7.0-1.x86_64.rpm
-
-The command above will install the ``/etc/yum.repos.d/crate.repo`` package
-repository configuration file.
+    [cratedb-ce-testing]
+    name=CrateDB RPM package repository - $basearch - Testing
+    baseurl=https://cdn.crate.io/downloads/yum/testing/7/$basearch
+    enabled=0
+    gpgcheck=1
+    gpgkey=https://cdn.crate.io/downloads/yum/RPM-GPG-KEY-crate
+    autorefresh=1
+    type=rpm-md
 
 .. NOTE::
 
-    CrateDB provides both *stable release* and *testing release* channels. You
-    can read more about the `release workflow`_.
+    The configured repository is disabled by default. This eliminates the
+    possibility of accidentally upgrading CrateDB when upgrading the rest
+    of the system. Each install or upgrade command must explicitly enable
+    the repository as indicated in the sample installation command below.
 
-    By default, yum_ (Red Hat's package manager) will use the stable
-    repository. This is because the testing repository is disabled.
-    If you would like to enable the testing repository, edit the ``crate.repo``
-    file and set ``enabled=1`` within the ``[crate-testing]`` section.
+CrateDB provides both *stable release* and *testing release* channels. You
+can read more about the `release workflow`_.
 
 
 Install CrateDB
@@ -50,7 +57,13 @@ Install CrateDB
 
 With everything set up, you can install CrateDB::
 
-    sudo yum install crate
+    sudo dnf install --enablerepo=cratedb-ce-stable crate
+
+.. TIP::
+
+    On older Red Hat and CentOS installations, please use the ``yum`` command
+    instead of ``dnf``. On SUSE based installations, please use the ``zypper``
+    command.
 
 
 Configure CrateDB
@@ -60,10 +73,25 @@ Please visit the :ref:`install-configure` documentation section to learn
 about the location and meaning of CrateDB's configuration files.
 
 
+Trust signing key
+=================
+
+In order to trust the package signing key upfront, before being prompted
+to do it on the first installation of CrateDB, you can also import it
+into your repository keyring, like that::
+
+    # Install prerequisites.
+    yum install sudo
+
+    # Import the public GPG key for verifying the package signatures.
+    sudo rpm --import https://cdn.crate.io/downloads/yum/RPM-GPG-KEY-crate
+
 
 .. include:: _control-linux.rst
 .. include:: _post-install.rst
 
+.. _DNF: https://en.wikipedia.org/wiki/DNF_(software)
 .. _release workflow: https://github.com/crate/crate/blob/master/devs/docs/release.rst
 .. _RPM: https://en.wikipedia.org/wiki/RPM_Package_Manager
 .. _YUM: https://en.wikipedia.org/wiki/Yum_(software)
+.. _ZYpp: https://en.wikipedia.org/wiki/ZYpp


docs/install/redhat.rst
```diff
@@ -28,11 +28,27 @@ from the CrateDB package repository::
...
+``/etc/zypp/repos.d/`` directory for OpenSuSE based distributions, containing::
+
+    [cratedb-stable]
+    name=CrateDB RPM package repository - $basearch - Stable
+    baseurl=https://cdn.crate.io/downloads/yum/7/$basearch
+    enabled=1
+    gpgcheck=1
+    gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-crate
+    autorefresh=1
+    type=rpm-md
```
amotl, line 42: Those two parameters have not been part of the previous package repository configuration. They are now inherited from the advised package repository configuration [like ES is doing it](https://www.elastic.co/guide/en/elasticsearch/reference/current/rpm.html#rpm-repo).

We did not research about their meaning yet, please advise correspondingly if you think they should be removed again.

docs/install/redhat.rst
```diff
@@ -28,11 +28,27 @@ from the CrateDB package repository::
...
-
-The command above will install the ``/etc/yum.repos.d/crate.repo`` package
-repository configuration file.
+To register with the CrateDB package repository, create a file called ``cratedb.repo``
+in the ``/etc/yum.repos.d/`` directory for RedHat based distributions, or in the
+``/etc/zypp/repos.d/`` directory for OpenSuSE based distributions, containing::
+
+    [cratedb-stable]
+    name=CrateDB RPM package repository - $basearch - Stable
+    baseurl=https://cdn.crate.io/downloads/yum/7/$basearch
```
amotl, line 37: ES doesn't use `$basearch`. Maybe we should also get rid of it?

docs/install/redhat.rst
```diff
@@ -28,11 +28,27 @@ from the CrateDB package repository::
     # Import the public GPG key for verifying the package signatures.
     sudo rpm --import https://cdn.crate.io/downloads/yum/RPM-GPG-KEY-crate
```
amotl, line 29: Hopefully, this command will drop the key at `/etc/pki/rpm-gpg/RPM-GPG-KEY-crate`, to match the repository configuration below? Actually, I did not validate the changes yet. Maybe someone of you can spare a few minutes?

docs/install/redhat.rst
```diff
@@ -28,11 +28,27 @@ from the CrateDB package repository::
...
-    sudo rpm -Uvh https://cdn.crate.io/downloads/yum/7/x86_64/crate-release-7.0-1.x86_64.rpm
-
-The command above will install the ``/etc/yum.repos.d/crate.repo`` package
-repository configuration file.
+To register with the CrateDB package repository, create a file called ``cratedb.repo``
+in the ``/etc/yum.repos.d/`` directory for RedHat based distributions, or in the
+``/etc/zypp/repos.d/`` directory for OpenSuSE based distributions, containing::
+
+    [cratedb-stable]
+    name=CrateDB RPM package repository - $basearch - Stable
```
amotl, line 36: Using "CrateDB" and "cratedb" now instead of "CRATE"/"crate" like before. Also using an explicit label `cratedb-stable` for designating the stable package repository.

docs/install/redhat.rst
hammerhead: It works both with and without for me. Not sure which one is more standard-compliant, but the less the better?

docs/install/redhat.rst
hammerhead: It was not stored under that location on Amazon Linux 2023, but everything still worked fine, with no warnings or errors. Potentially yum automatically knew where keys were stored and didn't attempt to check the manually specified path? Not sure, other repositories are using that path also:

```
[root@ip-172-31-26-116 ec2-user]# cat /etc/yum.repos.d/amazonlinux.repo 
[amazonlinux]
name=Amazon Linux 2023 repository
mirrorlist=https://al2023-repos-$awsregion-de612dc2.s3.dualstack.$awsregion.$awsdomain/core/mirrors/$releasever/$basearch/mirror.list
priority=10
enabled=1
repo_gpgcheck=0
type=rpm
gpgcheck=1
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-amazon-linux-2023

[amazonlinux-source]
name=Amazon Linux 2023 repository - Source packages
mirrorlist=https://al2023-repos-$awsregion-de612dc2.s3.dualstack.$awsregion.$awsdomain/core/mirrors/$releasever/SRPMS/mirror.list
enabled=0
repo_gpgcheck=0
type=rpm
gpgcheck=1
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-amazon-linux-2023
metadata_expire=6h

[amazonlinux-debuginfo]
name=Amazon Linux 2023 repository - Debug
mirrorlist=https://al2023-repos-$awsregion-de612dc2.s3.dualstack.$awsregion.$awsdomain/core/mirrors/$releasever/debuginfo/$basearch/mirror.list
enabled=0
repo_gpgcheck=0
type=rpm
gpgcheck=1
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-amazon-linux-2023
metadata_expire=6h
[root@ip-172-31-26-116 ec2-user]# ls /etc/pki/rpm-gpg/
RPM-GPG-KEY-amazon-linux-2022  RPM-GPG-KEY-amazon-linux-2023
[root@ip-172-31-26-116 ec2-user]# rpm --import https://cdn.crate.io/downloads/yum/RPM-GPG-KEY-crate 
[root@ip-172-31-26-116 ec2-user]# ls /etc/pki/rpm-gpg/
RPM-GPG-KEY-amazon-linux-2022  RPM-GPG-KEY-amazon-linux-2023
```

WalBeh (commented):
I did recently install cratedb on a fresh Amazon Linux host and run into this issues:
- Could install the `rpm` only with the `--nodeps`
```ansible
  - name: Import the GPG key for CrateDB
    tags: crate
    ansible.builtin.rpm_key:
      state: present
      key: https://cdn.crate.io/downloads/yum/RPM-GPG-KEY-crate

  - name: Register with the CrateDB package repository
    tags: crate
    ansible.builtin.command: rpm  -Uvh https://cdn.crate.io/downloads/yum/7/x86_64/crate-release-7.0-1.x86_64.rpm --nodeps
    ignore_errors: yes
```
- After `rpm` setup this was logged (enforced bootstrap checks) `Apr 05 15:54:37 c0 crate[8233]: [2024-04-05T15:54:37,464][INFO ][o.e.b.BootstrapChecks    ] [c0] bound or publishing to a non-loopback address, enforcing bootstrap checks
Apr 05 15:54:37 c0 crate[8233]: ERROR: [1] bootstrap checks failed`
- After that I added this to the playbook:

```
  - name: set limits for cratedb
    tags: crate, limits
    ansible.builtin.blockinfile:
      path: /etc/security/limits.conf
      block: |
        crate soft nofile unlimited
        crate hard nofile unlimited

        crate soft memlock unlimited
        crate hard memlock unlimited

        crate soft nproc 4096
        crate hard nproc 4096

        crate soft as unlimited
        crate hard as unlimited
      marker: "# {mark} ANSIBLE MANAGED BLOCK cratedb-limits"

  - name: Set sysctl parameters
    tags: crate, sysctl
    ansible.builtin.sysctl:
      name: "{{ item.name }}"
      value: "{{ item.value }}"
      state: present
      reload: yes
    loop:
      - { name: 'vm.max_map_count', value: '262144' }
      
  - name: Set ulimit
    tags: crate, sysctl
    ansible.builtin.pam_limits:
      domain: '*'
      limit_type: soft
      limit_item: nofile
      value: 65535

  - name: Install CrateDB
    tags: crate
    dnf:
      name: crate
      state: present
    become: yes
```

docs/install/redhat.rst
amotl: > The `gpgkey` setting also supports a URL. So we can skip the manual `rpm --import` and just set `gpgkey=https://cdn.crate.io/downloads/yum/RPM-GPG-KEY-crate`.

That sounds sweet. If there are no other objections, I will change it like this.

amotl:
Dear @WalBeh,

thanks for your suggestions. Needing to install `crate-release-7.0-1.x86_64.rpm` will be a thing of the past. Whether the CrateDB package itself should drop any kind of "limits" configuration file, for example into `/etc/security/limits.d`, would be a different matter, no?

Manipulating `/etc/security/limits.conf` directly is not advisable in any case, right? Is there anything else to consider? If the "limits" configuration is not part of CrateDB, should it be added to the installation documentation instead? If so, we also need to consider this for the other, non-RPM, installation methods?

With kind regards,
Andreas.

docs/install/redhat.rst
amotl: I am also for "the less the better" in general, yet quoting the documentation about `$basearch`, and it looks pretty standard amongst Red Hat at least, and may help if we ever release packages for ARM, for example?

> You can use `$basearch` to reference the base architecture of the system. For example, i686 machines have a base architecture of `i386`, and AMD64 and Intel 64 machines have a base architecture of `x86_64`.
>
> -- https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/sec-using_yum_variables

seut: Yes I think as soon we release arm packages this is needed or? So lets keep it..

docs/install/redhat.rst
```diff
@@ -28,11 +28,27 @@ from the CrateDB package repository::
...
+    baseurl=https://cdn.crate.io/downloads/yum/7/$basearch
+    enabled=1
```
seut, line 38: Do we want to enable it by default? ES doesn't for a reason, see the note at https://www.elastic.co/guide/en/elasticsearch/reference/current/rpm.html#rpm-repo.

amotl: That's the rationale provided by ES:

> The configured repository is disabled by default. This eliminates the possibility of accidentally upgrading `elasticsearch` when upgrading the rest of the system. Each install or upgrade command must explicitly enable the repository as indicated in the sample commands above.

For us, it would be a change I guess, because it also was `enabled=1` before. I will be happy to follow ES's suggestions however, like you are proposing.

docs/install/redhat.rst
amotl: Just to share my observations, probing the installation on CentOS 9 Stream, and AlmaLinux 9.

```
racker --verbose run -it --rm almalinux:9 /bin/bash
racker --verbose run -it --rm quay.io/centos/centos:stream9 /bin/bash
```

I missed to check on AlmaLinux, but on CentOS, the `gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-crate` setting does not have any kind of relevance any longer. When using `rpm --import` to import a public GPG key, it will no longer be located into `/etc/pki/rpm-gpg` at all, apparently.

amotl:
Dear @WalBeh,

on my [recent probings](https://github.com/crate/cratedb-guide/pull/69#discussion_r1565939653), I may have received this message:

```
/var/tmp/rpm-tmp.RXQENZ: line 4: sysctl: command not found

Warning: unable to set vm.max_map_count; is this an OpenVZ
instance? If so, it is highly recommended that you set
vm.max_map_count to 262144 on the host.
```

I did not turn too much attention to it, because, well, I was testing it on an _OS container instance_, where I accepted that `sysctl` may not be available. I probably need to validate this once more on a _real VM_.

However, I wanted to ask if that reflects "your issue"? At [your comment](https://github.com/crate/cratedb-guide/pull/69#pullrequestreview-1993450228), you did not exactly report about the problem, only the solution you applied.

Is it related to sysctl settings in any way, because you have been defining them on behalf of the Ansible recipe then?

Cheers,
Andreas

docs/install/redhat.rst
amotl: On our tests, they didn't cause any harm, so we just followed ES here, and included them to the new configuration

docs/install/redhat.rst
amotl: The patch has been amended to do it like advised. It seems to work well. Thanks.

docs/install/redhat.rst
amotl: I've also followed this advice, and amended the patch accordingly. Thanks.

hammerhead (approved):

PR #53, opened 2024-02-28
Evolve "Features" and "Application Domains" sections

amotl:
## About
This patch intends to **improve the actual usability of the non-reference documentation**, in terms of https://github.com/crate/cratedb-guide/labels/guidance. It is mostly based on https://github.com/crate/cratedb-guide/labels/cross%20linking, but also adds some https://github.com/crate/cratedb-guide/labels/new%20content where there have been gaps on properly outlining features and use cases, and in particular, where the documentation failed to answer the question "What can be done with CrateDB?" properly and quickly.

The patch also tries to provide a clear _information **architecture**_ to this realm of the documentation, i.e. concise categories where content should go, through the _physical entities of **folders**_, and correspondingly aligned _logical entities of documentation **sections**_, using a reasonable »_naming things_« approach. [^naming-things]

The improvements mostly focus on the "Features" and "Application Domains" sections, because the third one, "Integrations", already inherited its layout and shape from previous curations and improvements over at [Drivers and Integrations](https://cratedb.com/docs/clients/), so it doesn't need to recieve so much love on this iteration, we will come to that later once more.

[^naming-things]: Naming things is important. Please provide feedback where you would like to see a different wording. Thank you.

## Preview
- https://cratedb-guide--53.org.readthedocs.build/feature/
- https://cratedb-guide--53.org.readthedocs.build/domain/

## Thoughts

- It was clearly a failure to let engineers of other expertises manage the conveyance of relevant information details about the matters of **feature and use case presentation** on "the homepage". Up until now, despite receiving so many iterations over so many years, it does not manage to illustrate CrateDB's features and use cases concisely and thoroughly. This patch aims to remedy that.

- It was clearly a failure to separate the technical writing workbenches into "howtos" vs. "tutorials", as also elaborated elsewhere. Both terms/labels/categories are not tangible options to build an information architecture on. Previous refactoring patches already remedied this physically, this patch now concludes the improvement by evolving the semantic backbone into "features" vs. "application domains" vs. "integrations" instead.

- Contrary to what has been observed elsewhere, the jargon abstains from the typical [LLM yapping](https://twitter.com/vboykis/status/1751307750712156662) seen on many other spots these days, by just not using it.

[...]

PR #70, opened 2024-04-16
Move Cloud tutorials to Guide

matkuliak:
We're moving the tutorials from Cloud here

## References

- https://github.com/crate/cloud-docs/issues/68
- https://github.com/crate/infrastructure/pull/3149

### DIFF
--- a/docs/domain/document/index.md
+++ b/docs/domain/document/index.md
@@ -1,4 +1,5 @@
 (document)=
+(object)=
 # Document Store
 
 Learn how to efficiently store JSON or other structured data, also nested, and
@@ -8,7 +9,7 @@ Storing documents in CrateDB provides the same development convenience like the
 document-oriented storage layer of Lotus Notes / Domino, CouchDB, MongoDB, and
 PostgreSQL's `JSON(B)` types.
 
-- [](#object-marketing)
+- [](#objects-basics)
 - [Unleashing the Power of Nested Data: Ingesting and Querying JSON Documents with SQL]
 
 
@@ -18,5 +19,5 @@ PostgreSQL's `JSON(B)` types.
 :maxdepth: 1
 :hidden:
 
-object
+objects-hands-on
 ```
--- a/docs/domain/document/index.md
+++ b/docs/domain/document/index.md
@@ -1,5 +1,4 @@
 (document)=
-(object)=
 # Document Store
 
 Learn how to efficiently store JSON or other structured data, also nested, and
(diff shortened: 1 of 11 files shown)


docs/domain/search/index.md
amotl, line 1: This tutorial is clearly just a tutorial, and should not replace the main "index" page about "Search" within the "Application Domains" section.

Instead, it should be similar like you did it at the "Time Series" section: Have separate pages about the tutorials, and link them into the index file.

docs/domain/timeseries/time-series-advanced.md
amotl, line 1: I'd call that file `timeseries-and-metadata.md`. Please also use a similar label as target reference on the header, maybe `timeseries-objects`.

The idea is to use something that reasonably qualifies the content. In comparison to `objects-basic`, which conveys that it is about basics with OBJECTs, `advanced-guide` does not, because nobody knows what the "advanced" is relating to, and "guide" just adds redundancy.

docs/domain/timeseries/time-series.md
amotl, line 1: I'd call that file `timeseries-querying.md`. Please also use a similar label as target reference on the header.

docs/domain/document/index.md
```diff
@@ -1,5 +1,4 @@
 (document)=
-(object)=
```
amotl, line 2: Did you drop the `object` reference by accident?

docs/domain/document/object.md
amotl, line 1: I'd call that file `objects-hands-on.md` or such, instead of using a super canonical name `object`.

docs/domain/document/index.md
amotl: Ah, it is probably because there has been introduced a file called `object.md`, which led to collisions, right? Well, if we rename the file, this reference can stay.

docs/domain/document/object.md
```diff
@@ -0,0 +1,128 @@
+(object-marketing)=
```
amotl, line 1: Maybe?
```suggestion
(tutorial-objects)=
```
```suggestion
(object-basics)=
```

amotl (changes requested):
Thank you for carrying this over. I've added a few suggestions about »naming things«.

matkuliak:
> I've added a few suggestions about »naming things«.

Edited, thank you for good suggestions.

amotl (approved):
Thank you, approving. Currently, it does not make sense to employ the redirects, because GH-53 will change the structure: Both "document" and "search" will go to the "Features" section, so redirects would need to be changed for another time.

I think I wanted to make you aware of this, so you could have derived your feature branch directly from `amo/timeseries-more`. Apologies if I missed that, so please go ahead and just merge, in order to let me do the refactoring afterwards.

matkuliak:
Ah, I forgot about that one. 
Your #53 was started before this, please go ahead. Happy to wait and shuffle this around when it fits

PR #73, opened 2024-05-06
crate-docs-theme==0.32.0.dev14 + Remove `canonical_url_path` theme option

amotl:
Preview.

PR #74, opened 2024-05-06
Theme: Remove `canonical_url_path` theme option

amotl:
## About

Most recent `crate-docs-theme` removes the `canonical_url_path` theme option.

## References

- Carry over a fix from preview PR GH-73.


msbt (approved):
lgtm!

PR #95, opened 2024-07-08
Troubleshooting: Trim page about "Using `jcmd` with CrateDB on Docker"

amotl:
## Update

The procedure to use jcmd to attach to the CrateDB process running inside a container, by needing to prefix a `chroot` command, is apparently no longer needed.

    chroot --userspec=1000 /

Instead, it is sufficient to use `--user=crate` on the `docker run` command today, apparently.

    docker exec -it --user=crate cratedb /bin/bash
    /crate/jdk/bin/jcmd 1 VM.version

### DIFF
--- a/docs/admin/troubleshooting/jcmd/docker.rst
+++ b/docs/admin/troubleshooting/jcmd/docker.rst
@@ -32,57 +32,18 @@ Run ``jcmd`` inside container
 The commands below use ``docker``. In the same spirit, you can also use
 ``podman``.
 
-.. rubric:: Problem
+.. rubric:: Connect with user ``crate``
 
-After starting a ``cratedb`` Docker container,
-
-.. code-block:: console
-
-    docker run --rm -it --name=cratedb \
-        --publish=4200:4200 --publish=5432:5432 \
-        --env=CRATE_HEAP_SIZE=2g crate/crate:nightly \
-        -Cdiscovery.type=single-node
-
-you can also run ``jcmd`` inside the container.
+When aiming to invoke ``jcmd`` inside the container, please make sure to
+impersonate as the ``crate`` user.
 
 .. code-block:: console
 
-   $ docker exec -ti cratedb /bin/bash
-   [root@cratedb data]# /crate/jdk/bin/jcmd -l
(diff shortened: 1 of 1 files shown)


docs/admin/troubleshooting/jcmd/docker.rst
```diff
@@ -269,5 +243,21 @@ Java Flight Recorder (JFR)
    is not "blown up".
 
 
+Notes
+=====
+
+On earlier versions of the CrateDB OCI image, you needed to use ``chroot`` to
```
mfussenegger, line 249: Isn't this up to the docker version instead of the image? `--user` is a docker command option

docs/admin/troubleshooting/jcmd/docker.rst
```diff
@@ -32,57 +32,18 @@ Run ``jcmd`` inside container
 The commands below use ``docker``. In the same spirit, you can also use
```
mfussenegger, line 32: `podman` has no `--user` so looks like this no longer applies?

docs/admin/troubleshooting/jcmd/docker.rst
amotl: Thanks! It also works when using `su` instead of `chroot`. I will amend the patch correspondingly.

docs/admin/troubleshooting/jcmd/docker.rst
amotl: Bummer. Dito: I will use `su` instead of proprietary options.

amotl: It looks like both `podman run` and `podman exec` also accept the `--user` option?

-- https://docs.podman.io/en/stable/markdown/podman-run.1.html#user-u-user-group
-- https://docs.podman.io/en/stable/markdown/podman-exec.1.html#user-u-user-group

docs/admin/troubleshooting/jcmd/docker.rst
amotl: I've tried to improve the wording with f60c48328.

mfussenegger (approved):

Issue #101, opened 2024-07-20
Backlog for "Features" section


amotl:
## About
Coming from GH-53, there are a few backlog items, and there will be more.

## Details

Feature / Search: Improve layout
- https://github.com/crate/cratedb-guide/pull/53#discussion_r1529344885 by @surister

Feature / Vector: Add example using euclidean distance function
- https://github.com/crate/cratedb-guide/pull/53#discussion_r1685430878 by @amotl

Feature / Geo: What about querying with "donut" shapes?
- https://github.com/crate/cratedb-guide/pull/53#pullrequestreview-2160610407 by @seut

amotl:
## Discussion about Indexes

> Hi guys, I'd like to understand better CrateDB indexes, I have a few questions, thanks in advance.

[1] https://cratedb.com/blog/indexing-and-storage-in-cratedb
[2] https://rockset.com/blog/converged-indexing-the-secret-sauce-behind-rocksets-fast-queries/

### Questions
1. In [1] we say that "Inverted Indexes for text values, BKD-Trees for numeric values, and Doc Values." Is this still accurate in 2024 or do we implement any other data structure?
2. If I understood well in [1], given a column with default index (plain), do we build into an inverted index + columnar (doc values) for text values, for example, or is doc values reserved only for things like objects/arrays? Maybe another way of putting the question is, do we only use one index data structure per datatype/column, or do we apply more than one and then maybe choose to query one or another depending of the user's query in the optimizer?
3. In [1] "...new documents are added to the existing index, they are added to the next segment ...the system may decide to merge some segments ...adding a new document does not require rebuilding the index structure" Is this the reason why our index-all strategy does not affect insert performance as much as other databases when you manually set to build index in several columns? or is there any other strategy in place to mitigate the overhead of indexing every column?
4. In [2] they defined converged index as: row (LSM trees) + columnar + search (posting lists), how fair would be to say that we also 'implement' converged index in your opinion?
5. In [1] "...CrateDB implements Column Store based on Doc Values in Lucene" Does this mean that we just use Lucene's DocValues or do we wrote our own based/inspired on it?

### Answers
1. This is still accurate.
2. Everything gets doc values; numeric and geo types also get a BKD index; text and index fields also get postings lists.  And yes, which index structure is used depends on the query.
[...]

PR #106, opened 2024-07-30
Feature/Search: Reorganize section, and add content about hybrid search

amotl:
## About
Inspired by [Doing Hybrid Search in CrateDB](https://cratedb.com/blog/hybrid-search-in-cratedb) (thanks @surister!), this patch intends to reoragnize the "Search" section, in order to make its structure strong enough to cover the discrimination between Full-Text Search, Vector Search, and Hybrid Search well.

## Preview
_Rendering._


docs/feature/search/vector.md
```diff
@@ -0,0 +1,42 @@
...
+## Learn
+
+To learn more about vector search, please visit the corresponding page about
+[](#vector-store).
+
+
+:::{todo}
+Bring page into the same shape like the others in this section.
+Maybe just move the [](#vector-store) page here without further ado.
+:::
```
amotl, line 37: Better repurpose and dissolve the main-level [Vector Store](https://cratedb.com/docs/guide/feature/vector/) page completely, and just move it here, in order to provide a better narrative and enhanced guidance. wdyt?

surister (approved):
Brilliant, I love the idea and look forward putting in more content, great stuff

surister:
At the beginning you talk about how vectors is not enough hence we need to mix with bm25, this is very well written in the description of https://haystackconf.com/us2023/talk-16/, maybe it can serve as an inspiration?

amotl:
> Brilliant, I love the idea and look forward putting in more content, great stuff.

Thanks a stack, appreciate it.

> See [...] for other sources of inspiration.

I've adjusted my OP by adding a few inspirations I had slightly used for conceiving the first sketch of this patch. I was sure you would provide much better resources than coming from my quick research. Thanks!

I will be so happy to accept well rewritten paragraphs on this PR, based on the material you are suggesting, if you can afford a few cycles. Every chunk counts!

Otherwise, we will probably need to add it to the backlog summary at GH-101, because I need to take care about other obligations.

docs/feature/search/vector.md
amotl: What do you think about this, @surister? In my impression, looking at it from a fresh perspective, listing "Vector Store" as a main feature does not make too much sense, because you will mostly use it for "SEARCHing".

In this spirit, I am planning to move that page to the revamped /search/ section, in order to make the trio of (fts, vector, hybrid) complete. Do you agree?

amotl: Hm. Thinking about it once more, the Vector Store page might still make sense to have "Vector" enumerated on the top menu level in that section. Otherwise, it might make people feel it is missing here.

Maybe still keep that page, but switch the content around?

amotl:
> [improve] about how vectors is not enough hence we need to mix with bm25

I've diverted this to https://github.com/crate/cratedb-guide/issues/101#issuecomment-2258857679. Thanks!

docs/feature/search/vector.md
amotl: Hi again. I decided to [refactor mercilessly](http://www.extremeprogramming.org/rules/refactor.html), and dissolved both the "Vector Store" and the "Geospatial Data" page, refactoring them into "Vector Search" and "Geospatial Search", now making up a quartett together with the newly added "Hybrid Search" page.

image

amotl: That other patch adds relevant redirects to compensate for refactored pages now available on a different location.

- https://github.com/crate/infrastructure/pull/3197

PR #112, opened 2024-08-01
Theme: Update to crate-docs-theme==0.34.0.dev5

amotl:
Increases toctree maxdepth to 4.


amotl:
> **Also increases toctree maxdepth to 4**. Before, it was 3.

It certainly makes a difference, for example here. We will have to decide if we want it or not.

- Now: https://cratedb.com/docs/guide/performance/inserts/methods.html
- Future? https://cratedb-guide--112.org.readthedocs.build/performance/inserts/methods.html

### Before/After
A picture says...

image image


/cc @msbt, @matkuliak, @simonprickett, @karynzv, @hammerhead, @hlcianfagna, @surister, @proddata

matkuliak:
In general im for more depth. But the shared example is a bit confusing, at least to me.
The additional items in primary navigation aren't pages but headers/titles. So it's showing the same thing as the right-hand menu, but then lacking the depth for h2 headers.

amotl:
When there are no strong objections against it, I will be merging it. The CrateDB Guide has been used as a canary to run the modernized documentation theme on production as a canary.

PR #129, opened 2024-10-25
Add position and split to list of functions equivalence 

hlcianfagna:

This add equivalences for the position and split scalar functions


amotl (approved):
Thank you!

/cc @simonprickett, @kneth, just for the purpose of signalling how many small updates can make a difference.

NB: The GitHub web editor is good enough these days to do _small edits_ "inline and online" without even bothering to go into any other repository-, filesystem-, terminal-, and ui-interactions. Larger edits and authoring processes certainly needs professional tooling, that's why I also would like to take the opportunity to make our documentation sandbox subsystem (`make dev`) work well on Windows, to satisfy needs of @hlcianfagna.

PR #135, opened 2024-11-14
Feature/Connectivity: Rework page

amotl:
## Problem
The [Feature » Connectivity](https://cratedb.com/docs/guide/feature/connectivity/) page wasn't in a good shape yet. @kneth discovered that recently.

## Solution
This patch intends to converge the page into the same layout/shape/style like all the others in the "All Features" section, using the same compact backbone made of "Intro", "Synopsis", and "Learn" subsections.

## Preview
https://cratedb-guide--135.org.readthedocs.build/feature/connectivity/

## Review
- _Please visit inline comments._
- We believe the "tabs" approach will not scale, so we diverted using the "dropdowns" approach, currently showing both variants. Please advise.

## Details
This refactors a bit of content from a few sibling pages in crate-clients-tools. [^1]
- https://github.com/crate/crate-clients-tools/tree/main/docs/connect
- https://cratedb.com/docs/crate/clients-tools/en/latest/connect/

Demonstrate two different variants for single-page client-/driver-snippets presentation, one using [tabs](https://sphinx-design.readthedocs.io/en/latest/tabs.html), and another one using [dropdowns](https://sphinx-design.readthedocs.io/en/latest/dropdowns.html).

[^1]: crate-clients-tools might be dissolved in the future, in order to reduce maintenance overhead, so it's a good idea to migrate content gradually.

### DIFF
--- a/docs/feature/connectivity/index.md
+++ b/docs/feature/connectivity/index.md
@@ -1,85 +1,149 @@
 (connect)=
 (connectivity)=
 
-
 # Connectivity
 
-All the CrateDB connectivity options at a glance.
+:::{include} /_include/links.md
+:::
+
+:::::{grid}
+:padding: 0
+
+::::{grid-item}
+:class: rubric-slimmer
+:columns: auto 9 9 9
+
+
+:::{rubric} Overview
+:::
+CrateDB connectivity options at a glance.
 
 You have a variety of options to connect to CrateDB, and to integrate it with
 off-the-shelve, 3rd-party, open-source, and proprietary applications.
 
-{tags-primary}`Connect`
-{tags-primary}`Import`
-{tags-primary}`Export`
-{tags-primary}`Extract`
-{tags-primary}`Load`
-{tags-primary}`ETL`
-
-
-## Interfaces and Protocols
-
-CrateDB supports both the [HTTP protocol] and the [PostgreSQL wire protocol],
+:::{rubric} About
+:::
+CrateDB supports both the HTTP protocol and the PostgreSQL wire protocol,
 which ensures that many clients that work with PostgreSQL, will also work with
 CrateDB.
 
-Through corresponding drivers, CrateDB is compatible with [ODBC],
-[JDBC], and other database API specifications.
-By supporting [SQL], CrateDB is compatible with many standard database
+Through corresponding drivers, CrateDB is compatible with JDBC, ODBC,
+and other database API specifications.
+By supporting SQL, CrateDB is compatible with many standard database
 environments out of the box.
 
-- [CrateDB HTTP interface]
-- [CrateDB PostgreSQL interface]
-- [CrateDB SQL protocol]
-
-To learn more, please refer to the documentation sections about supported
-client drivers, libraries, and frameworks, and corresponding tutorials.
-
-
-## Drivers and Integrations
+:::{rubric} Details
+:::
 
 CrateDB provides plenty of connectivity options with database drivers,
 applications, and frameworks, in order to get time series data in and
 out of CrateDB, and to connect to other applications.
 
-- [Drivers and Integrations]
-- [](inv:crate-clients-tools#connect)
-- [](inv:crate-clients-tools#df)
-- [](inv:crate-clients-tools#etl)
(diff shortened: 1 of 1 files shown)


docs/feature/connectivity/index.md
```diff
@@ -1,85 +1,608 @@
...
+```shell
+docker run --rm -it --publish=4200:4200 --publish=5432:5432 crate/crate:nightly
+```
+
+
+::::::{tab-set}
+
+:::::{tab-item} C# / Npgsql
+For more information see [Npgsql] documentation.
+
```
amotl, line 88: That's the "tabbed" variant. We think it will not scale, and nesting tabs is also debatable.

docs/feature/connectivity/index.md
```diff
@@ -1,85 +1,608 @@
...
-:::{rubric} Community Tutorials
+:::::{dropdown} PHP
+
+:::{rubric} HTTP
+:::
+
+::::{dropdown} PDO_CRATEDB
+The PHP Data Objects (PDO) is a standard PHP extension that defines a common
+interface for accessing databases in PHP.
+For more information see [CrateDB PDO] documentation.
```
amotl, line 300: That's the "dropdown" variant, already using **nested dropdowns**, immediately applicable with PHP and Python for example, as there are many driver variants for them, across two families (HTTP vs. PostgreSQL), which stretches all axes of the system needed to represent those.

We are biased to use this variant, and drop the "tabbed" variant again. wdyt?

docs/feature/connectivity/index.md
```diff
@@ -1,85 +1,608 @@
...
-{tags-primary}`Connect`
-{tags-primary}`Import`
-{tags-primary}`Export`
-{tags-primary}`Extract`
-{tags-primary}`Load`
-{tags-primary}`ETL`
 
+:::{rubric} Overview
+:::
+All the CrateDB connectivity options at a glance.
```
kneth, line 19:
```suggestion
CrateDB connectivity options at a glance.
```

docs/feature/connectivity/index.md
```diff
@@ -1,85 +1,608 @@
...
+
+public class Main {
+    public static void main(String[] args) {
+        try {
+            Properties properties = new Properties();
+            properties.put("user", "crate");
+            properties.put("password", "");
+            properties.put("ssl", false);
+            Connection conn = DriverManager.getConnection(
+                "jdbc:crate://localhost:5432/",
```
kneth, line 191: please consider to add a comment here about the two different drivers

kneth (approved):
I have only minor suggestions

docs/feature/connectivity/index.md
amotl: The main question to you here is if we should continue with the dropdown variant. Thanks!

docs/feature/connectivity/index.md
kneth: I agree, on a small screen (14 in laptop) it is not an optimal solution

docs/feature/connectivity/index.md
kneth: If we go with the "tabbed" variant, I think we should pick 3-4 items only. So if we want to have all items, the "dropdown" variant might be better.

amotl: I tried to present both variants without many initial plans, and just related to a recent comment from your pen that you like "tabbed" approaches of information conveyance.

What do _you_ think how we should proceed on this particular matter?

amotl: Thanks for your suggestions. I've diverted corresponding updates into a separate patch that needs more love, in order to get this one merged.

- GH-138

docs/feature/connectivity/index.md
amotl: Thanks. Do you agree with that variant?

https://github.com/crate/cratedb-guide/blob/525483fa894f1da95bcffce5e1aa91fea30b6be5/docs/feature/connectivity/index.md?plain=1#L183-L197

- => See: GH-138. Please respond there to further improve that section. For example, we could add inline comments or a link to the relevant documentation page, when applicable.

PR #147, opened 2024-11-27
Memory Configuration: Add warnings about configuring heap size

henrikingo:

Add warning about the advice to configure heap dump. 

Basically:
heap dump is workload dependent
must test
may crash otherwise

### DIFF
--- a/docs/admin/memory.rst
+++ b/docs/admin/memory.rst
@@ -23,9 +23,16 @@ The right memory configuration depends on your workloads.
 Recommendation
 ==============
 
-A good starting point for the heap space is 25% of the system memory. However,
-it shouldn't be set below 1GB and not above 30.5GB, see the limits section
-below.
+A good starting point for the heap space is 25% of the system memory.
+
+Be careful when configuring this, as making the heap too large may lead to OutOfMemory errors,
+if the operating system cannot allocate memory for other needs, and on the other hand, making
+the heap too small may lead to OutOfMemory errors when new Java objects cannot be allocated
+inside the heap.
+
+In any case, it shouldn't be set below 1GB and not above 30.5GB,
+See the limits section below.
+
 
 Considerations
 ==============
@@ -40,6 +47,10 @@ Consider the following when determining the right value:
   collection times **increase** with a bigger heap, leading to **increased**
   latency. Therefore, your heap size shouldn't be too large.
 
+- The optimal heap size is workload dependent. Be sure to test carefully
+  if you want to change the default, as wrong values may cause CrateDB to crash
+  with OutOfMemory errors.
+
 - CrateDB needs to be able to hold all the structures required to process
   queries in memory. The biggest offender are intermediate or final result
   sets. Therefore, the heap region must be large enough to hold the biggest
@@ -49,6 +60,9 @@ Consider the following when determining the right value:
 - CrateDB uses `Memory mapped files`_ via `Lucene`_. This reduces the
   amount of heap space required and benefits from the file system cache.
 
+- Both to keep latency down, and to enable use of heap sizes above 32 GB,
+  it is also possible to use one of the newer garbage collectors, such as ZGC.
+  These are known to have worked, but aren't officially tested nor supported.
 
 .. _swap:
 


docs/admin/memory.rst
```diff
@@ -23,9 +23,16 @@ The right memory configuration depends on your workloads.
...
-below.
+A good starting point for the heap space is 25% of the system memory.
+
+Be careful when configuring this, as making the heap too large may lead to OutOfMemory errors,
+if the operating system cannot allocate memory for other needs, and on the other hand, making
+the heap too small may lead to OutOfMemory errors when new Java objects cannot be allocated
+inside the heap.
+
+In any case, it shouldn't be set below 1GB and not above 30.5GB,
+See the limits section below.
```
amotl, line 34: I would like to ask a question about the very detail of the upper memory limit recommendation, but not within the conversation on this PR. I am still using it to spawn a dedicated ticket from. Thanks.

docs/admin/memory.rst
```diff
@@ -49,6 +60,9 @@ Consider the following when determining the right value:
 - CrateDB uses `Memory mapped files`_ via `Lucene`_. This reduces the
   amount of heap space required and benefits from the file system cache.
 
+- Both to keep latency down, and to enable use of heap sizes above 32 GB,
+  it is also possible to use oneof the newer garbage collectors, such as ZGC.
```
amotl, line 64:
```suggestion
  it is also possible to use one of the newer garbage collectors, such as ZGC.
```

amotl (commented):
I like the update, thanks a stack for submitting it.

amotl:
@mfussenegger might be off for the week, so maybe @seut can also have a look and acknowledge?

seut (approved):

docs/admin/memory.rst
```diff
@@ -40,6 +47,10 @@ Consider the following when determining the right value:
   collection times **increase** with a bigger heap, leading to **increased**
   latency. Therefore, your heap size shouldn't be too large.
 
+- The optimal heap size is workload dependent. Be sure to test carefully
+  if you want to change the default, as wrong values may cause CrateDB to crash
+  with OutOfMemory errors.
+
```
mfussenegger, line 53: Isn't this just the same point as below?

> Therefore, the heap region must be large enough to hold the biggest
  result sets in memory. The size of the intermediate and final result sets
  depend entirely on the type of queries you are running.

Is the same as "optimal heap size is workload dependent" but with more details.

The "wrong values may cause CrateDB to crash with OutOfMemory errors." also
doesn't exactly raise confidence in the product - or give the impression that
we know what we're doing.

We should give concrete min values for both heap and off-heap, and after that
users should receive circuit breaker errors and not have their nodes crash.

docs/admin/memory.rst
```diff
@@ -23,9 +23,16 @@ The right memory configuration depends on your workloads.
...
 
-A good starting point for the heap space is 25% of the system memory. However,
-it shouldn't be set below 1GB and not above 30.5GB, see the limits section
-below.
+A good starting point for the heap space is 25% of the system memory.
+
+Be careful when configuring this, as making the heap too large may lead to OutOfMemory errors,
+if the operating system cannot allocate memory for other needs, and on the other hand, making
+the heap too small may lead to OutOfMemory errors when new Java objects cannot be allocated
+inside the heap.
```
mfussenegger, line 31: This is kinda duplicating the point further below?

docs/admin/memory.rst
```diff
@@ -49,6 +60,9 @@ Consider the following when determining the right value:
 - CrateDB uses `Memory mapped files`_ via `Lucene`_. This reduces the
   amount of heap space required and benefits from the file system cache.
 
+- Both to keep latency down, and to enable use of heap sizes above 32 GB,
+  it is also possible to use one of the newer garbage collectors, such as ZGC.
+  These are known to have worked, but aren't officially tested nor supported.
```
mfussenegger, line 65: I'd integrate this either in the first point that mentions G1GC and also include a link to GC configuration so users can learn how to change it. Or alternative put it into the `Recommendation` section further below under Memory Limits because it's kinda relevant to exceeding the 30.5 compressed oops space.

PR #151, opened 2024-11-28
dbt: Educate about `generate_schema_name` macro

amotl:
## About
While harvesting more ingredients from https://github.com/crate/cratedb-examples/pull/754, effectively originating at [Using dbt with CrateDB](https://community.cratedb.com/t/using-dbt-with-cratedb/1566), we are adding the custom `generate_schema_name` macro override to the documentation instead of slotting it into dbt-cratedb2, which we think is the right choice.

## Preview
_Rendering._

### DIFF
--- a/docs/integrate/dbt/index.md
+++ b/docs/integrate/dbt/index.md
@@ -86,6 +86,30 @@ options apply like outlined on the [dbt Postgres Setup] documentation
 page.
 
 
+## Usage
+
+### Custom Schemas
+By default, dbt writes the models into the schema you configured in your
+profile, but in some dbt projects you may need to write data into different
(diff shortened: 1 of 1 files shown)


hlcianfagna (approved):
It is true this is not CrateDB specific so I think your decision to have it here instead of inside `dbt-cratedb2` is the right choice. For context this was addressing an issue with materialization of VIEWs which we could reproduce with PostgreSQL and without any macro overrides, see https://discourse.getdbt.com/t/schema-name-is-not-correct/6816

amotl:
Thank you Hernan. I created a ticket about it, so we should investigate, in order to not introduce any regressions that have been resolved already.

- https://github.com/crate/dbt-cratedb2/issues/5

PR #167, opened 2024-12-17
Add REGEXP_LIKE to list of Rockset function equivalences

hlcianfagna:
## About

This add the equivalence for `REGEXP_LIKE` to the page which covers the translation of queries from Rockset

Ref https://docs.rockset.com/documentation/reference/regular-expression-functions


amotl:
Hi. Thanks for submitting this improvement. I guess it is safe to merge the update?

PR #173, opened 2025-01-09
Updated URL used to access example data.

simonprickett:
I've updated the URL used to access the example data to be the CrateDB one rather than direct access from GitHub.


kneth (approved):
Thank you

PR #182, opened 2025-03-04
About query optimization and considerations for scaling

hlcianfagna:
## About

Adding two new big sections about:

- Query Optimization 101
- Topics to watch out when productizing CrateDB

## Preview

- https://cratedb-guide--182.org.readthedocs.build/performance/optimization.html
- https://cratedb-guide--182.org.readthedocs.build/performance/productizing.html


docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+.. code:: text
+
+   SELECT *
+   FROM mytable
+   WHERE
+     CASE
+           WHEN $1 = 'ALL COUNTRIES' THEN true
+           WHEN $1 = mytable.country AND $2 = 'ALL CITIES' THEN true
+           ELSE $1 = mytable.country AND $2 = mytable.city
+     END;
```
amotl, line 261: Hi. I'd like to emphasize this spot, because it is where currently the Pygments SQL formatter fails.

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+
+can be rewritten as
+
+.. code:: text
+
+   SELECT *
+   FROM mytable
+   WHERE ($1 = 'ALL COUNTRIES')
+   OR ($1 = mytable.country AND $2 = 'ALL CITIES')
+   OR ($1 = mytable.country AND $2 = mytable.city)
```
amotl, line 271: That is another spot.

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+instead:
+
+.. code:: sql
+
+   SELECT customerid,
+     (SELECT SUM(order_amount)
+      FROM customer_orders
+      WHERE customer_orders.customerid = customers.customerid
+     ) AS total
+   FROM customers;
```
amotl, line 338: Contrary to the two samples above, this spot, including less advanced SQL, is well understood by Pygments.

amotl (commented):
Thanks a stack for submitting this patch. I didn't do a review, but I've outlined two spots to be able to link them into an upstream issue ticket on [Pygments](https://pypi.org/project/Pygments/).

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+****************************
+ Only sort data when needed
+****************************
+
+Indexing in CrateDB is optimized to support filtering and aggregations without
+requiring expensive defragmentation operations, but it is not optimized for
+sorting​.
+
+Maintaining a sorted index would slow down ingestion​ and Redshift and other
+analytical database systems like Cassandra make similar trade-offs​.
```
amotl, line 102:
```suggestion
Maintaining a sorted index would slow down ingestion, that's why​ other analytical
database systems like Cassandra and Redshift make similar trade-offs​.
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+
+This means that when an ``ORDER BY`` is requested the whole dataset comes in
+memory in a node and data is sorted there and then, hence it is important to not
+request ``ORDER BY`` operations when not actually needed, there is, of course,
+no problem sorting a few thousand rows in the final stage of a ``SELECT`` but we
+need to avoid requesting sort operations over millions of rows.
```
amotl, line 108:
```suggestion
This means that when an ``ORDER BY`` operation is requested, the whole dataset
needs to be loaded into the main memory on the relevant cluster node to be sorted.
That is why it is important to not request ``ORDER BY`` operations when not actually
needed, and most importantly, not on tables of large cardinalities without aggregating
records beforehand. On the other hand, of course it is no problem to sort a few
thousand rows in the final stage of a ``SELECT`` operation, but we need to avoid
requesting sort operations over millions of rows.
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+
+Consider leveraging filters and aggregations like ``max_by`` and ``min_by`` to
+get the desired results limiting the scope of ``ORDER BY`` operations or
+avoiding them altogether.
```
amotl, line 112:
```suggestion
limit the scope of ``ORDER BY`` operations, or
avoid them altogether when possible.
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+   ORDER BY reading_time DESC
+   LIMIT 10;
+
+.. _filter-with-array-expressions:
+
+***************************************************************************
+ Use filters with array expressions when filtering on the output of UNNEST
+***************************************************************************
+
+On denormalized data sets you may have records with an array of objects.
```
amotl, line 140:
```suggestion
On denormalized data sets, you may observe records including columns
storing arrays of objects.
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+
+You may want to unnest the array in a subquery or CTE and later filter on a
+property of the OBJECTs.
```
amotl, line 143:
```suggestion
You may want to unnest the array in a subquery or CTE and later filter on a
property of the OBJECTs.

The next statement will result in every row in the table (not filtered
with other conditions) being read and unnested, to check if it meets
the criteria on ``field1``.
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+   FROM (
+      SELECT UNNEST(my_array_of_objects) obj
+      FROM my_table
+   )
+   WHERE obj['field1'] = 1;
+
+Just written like that this will result in every row in the table (not filtered
+with other conditions) being read and unnested to check if it meets the criteria
+on ``field1``, but CrateDB can do a lot better than this if we add an additional
+condition like this:
```
amotl, line 157:
```suggestion
However, CrateDB can do a lot better than this if we add an additional condition
like this:
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+
+.. _format-as-last-step:
+
+******************************
+ Format output as a last step
+******************************
+
+In many cases, data may be stored in an efficient format but we want to
+transform it to make it more human-readable in the output of the query, we may
+use `scalar functions`_ such as ``date_format`` or ``timezone``.
```
amotl, line 180:
```suggestion
In many cases, data may be stored in an efficient format, but we want to
transform it to make it more human-readable in the output of the query.
To accomodate such situations, we may use `scalar functions`_ such as
``date_format`` or ``timezone``.
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+Sometimes queries apply these transformations in an intermediate step and later
+do further operations like filtering on the transformed values.
+
+The CrateDB optimizer is actually very good at seeing through many of these
+situations and still using indexes on the original data. But there is always a
+risk that something particular in the query prevents this from happening and we
+end up applying the transformation on thousands or millions of records that
+later will be discarded. So whenever makes sense we want to only apply these
+transformations when we have already worked out the final result set to be sent
+back to the client.
```
amotl, line 191:
```suggestion
CrateDB's query optimizer attempts to determine the most efficient way to
execute a given query by considering the possible query plans. Based on
the query scenario/situation, it is always aiming to use existing indexes on
the original data for maximum efficiency.

However, there is always a chance that some particular clause in the query
expression prevents the optimizer from selecting an optimal plan, ending up
applying the transformation on thousands or millions of records that later
would be discarded anyway. So, whenever it makes sense, we want to make
sure these transformations are only applied after the database has already
worked out the final result set to be sent back to the client.
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+     (SELECT SUM(order_amount)
+      FROM customer_orders
+      WHERE customer_orders.customerid = customers.customerid
+     ) AS total
+   FROM customers;
+
+.. _batch-operations:
+
+******************
+ Batch operations
```
amotl, line 343: Maybe call the headline like what's inside? Otherwise, "batch operations" will easily be read differently, like I was just doing it, expecting to see traditional database batched inserted operations instead.

```suggestion
 Batch approach
```

Maybe even extend the headline to say "Batch approach for `UPDATE` and `INSERT ... FROM ... SELECT` operations", to make it clearer, if that's not too long?

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+*****************************************************************************
+ Use staging tables for intermediate results if you are doing a lot of JOINs
+*****************************************************************************
+
+If you have many CTEs or VIEWs and need to JOIN these in some cases it can be
+effective to store the intermediate results from these into dedicated tables and
+then use these tables, while there is a cost in writing to disk and reading data
+back we can benefit from indexing and from giving the optimizer more
+straightforward execution plans that it can optimize for parallel execution
+using multiple nodes in the cluster.
```
amotl, line 404:
```suggestion
If you have many CTEs or VIEWs with a need to JOIN them, it can be benefical
to query them individually, store intermediate results into dedicated tables, and
then use these tables for JOINing.

While there is a cost in writing to disk and reading data back, the whole operation
can benefit from indexing and from giving the optimizer more straightforward
execution plans, to enable it optimizing for better parallel execution using
multiple cluster nodes.
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+   FROM customers
+
+.. _consider-generated-columns:
+
+****************************
+ Consider generated columns
+****************************
+
+If you frequently find yourself extracting information from fields and then
+using this extracted data on filters or aggregation it can be good to consider
```
amotl, line 432:
```suggestion
If you frequently find yourself extracting information from fields and then
using this extracted data on filters or aggregations, it can be good to consider
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+ Be mindful of UDFs, leverage them in the right contexts, but only in the right contexts
+*****************************************************************************************
+
+UDFs run on a javascript virtual machine and on a single thread, so they can
+have an impact on performance.
+
+Also once we pass a value through an UDF the engine will have to work with the
+results in memory and will not be able to leverage indexes on the underlying
+fields anymore so the general considerations about delaying formatting as much
+as possible apply.
```
amotl, line 451:
```suggestion
When using user-defined functions (UDFs), two important details relevant for
performance aspects need to be considered.

1. Once records or values will be processed by an UDF, the database engine will
   need to load results into memory, and will not be able to leverage indexes
   any longer. In this spirit, please apply the relevant general considerations
   about delaying formatting as much as possible.

2. UDFs run on a JavaScript virtual machine on a single thread, so they can have
   an impact on performance.
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+SQL.
+
+.. _positive-filters:
+
+*******************************************************************
+ Prefer positive filter expressions to negative filter expressions
+*******************************************************************
+
+Positive filter expressions can directly leverage indexing, sometimes the
+optimizer may be able to rewrite a negative expression to still use indexes but
```
amotl, line 463:
```suggestion
Positive filter expressions can directly leverage indexing. With negative
expressions, the optimizer **may** be able to rewrite it to still use indexes, but
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+   WHERE customerid > 3 AND status = 'active';
+
+.. _use-null-or-empty:
+
+******************************************************************************
+ Use the special null_or_empty function with OBJECTs and ARRAYs when relevant
+******************************************************************************
+
+CrateDB has a special scalar function called null_or_empty_ , using this in
+filter conditions against OBJECTs and ARRAYs is much faster that look for ``IS
```
amotl, line 495:
```suggestion
filter conditions against OBJECTs and ARRAYs is much faster than to look for ``IS
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+   FROM mytable
+   WHERE null_or_empty(array_column);
+
+.. _execution-plans:
+
+************************
+ Review execution plans
+************************
+
+If a query is slow but still completes in a certain amount of time we can use
```
amotl, line 520:
```suggestion
If a query is slow but still completes in a certain amount of time, we can use
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+`EXPLAIN ANALYZE`_ to get a detailed execution plan, the main thing to watch for
```
amotl, line 521:
```suggestion
`EXPLAIN ANALYZE`_ to get a detailed execution plan. The main thing to watch for
```

amotl (commented):
Hi Hernan. Thanks a stack for submitting this extensive page about CrateDB querying guidelines. While I've not reviewed the SQL statements, I've gone through the prose and added a few suggestions, just to use at your disposal.

Disclaimer: As I am not a native speaker, I might easily be wrong about many wording, syntactic, and punctuation details I've outlined. Please outright ignore or reject corresponding suggestions.

docs/performance/optimization.rst
amotl: Using the `plpgsql` lexer is the right choice here.

```shell
uv pip install pygments
```
```shell
pygmentize -L
```

amotl: Those are all lexers for SQL variants available in Pygments.

```shell
uv pip install pygments
```
```shell
pygmentize -L
```
```
pygmentize -L --json | jq '.lexers | with_entries( select(.key | test("sql"; "i") ))' | yq --input-format json
```
```yaml
GSQL:
  aliases:
    - gsql
  filenames:
    - '*.gsql'
  mimetypes: []
GoogleSQL:
  aliases:
    - googlesql
    - zetasql
  filenames:
    - '*.googlesql'
    - '*.googlesql.sql'
  mimetypes:
    - text/x-google-sql
    - text/x-google-sql-aux
MySQL:
  aliases:
    - mysql
  filenames: []
  mimetypes:
    - text/x-mysql
PL/pgSQL:
  aliases:
    - plpgsql
  filenames: []
  mimetypes:
    - text/x-plpgsql
PostgreSQL console (psql):
  aliases:
    - psql
    - postgresql-console
    - postgres-console
  filenames: []
  mimetypes:
    - text/x-postgresql-psql
PostgreSQL EXPLAIN dialect:
  aliases:
    - postgres-explain
  filenames:
    - '*.explain'
  mimetypes:
    - text/x-postgresql-explain
PostgreSQL SQL dialect:
  aliases:
    - postgresql
    - postgres
  filenames: []
  mimetypes:
    - text/x-postgresql
SQL+Jinja:
  aliases:
    - sql+jinja
  filenames:
    - '*.sql'
    - '*.sql.j2'
    - '*.sql.jinja2'
  mimetypes: []
SQL:
  aliases:
    - sql
  filenames:
    - '*.sql'
  mimetypes:
    - text/x-sql
sqlite3con:
  aliases:
    - sqlite3
  filenames:
    - '*.sqlite3-console'
  mimetypes:
    - text/x-sqlite3-console
Transact-SQL:
  aliases:
    - tsql
    - t-sql
  filenames:
    - '*.sql'
  mimetypes:
    - text/x-tsql
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+
+.. _use-null-or-empty:
+
+******************************************************************************
+ Use the special null_or_empty function with OBJECTs and ARRAYs when relevant
+******************************************************************************
+
+CrateDB has a special scalar function called null_or_empty_ , using this in
+filter conditions against OBJECTs and ARRAYs is much faster that look for ``IS
+NULL`` if accepting empty objects and arrays is acceptable.
```
amotl, line 496:
```suggestion
filter conditions against OBJECTs and ARRAYs is much faster than using
the ``IS NULL`` clause, if accepting empty objects and arrays is acceptable.
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,544 @@
...
+
+.. _minimise-result-sets:
+
+*************************
+ Avoid large result sets
+*************************
+
+Be aware of the number of rows you are returning in a ``SELECT`` query.
+Analytical databases, such as CrateDB, excel at processing large data sets and
+returning small to medium-sized result sets. Serializing, transporting them over
```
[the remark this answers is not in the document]
hlcianfagna: I need to credit @hammerhead  for this wording

docs/performance/optimization.rst
hlcianfagna: That worked, thank you

amotl (approved):
Acknlowedging, to unblock the merge if you need it to be published.
Thank you again! 💯

docs/performance/productizing.rst
```diff
@@ -0,0 +1,116 @@
+.. _topics-to-watch-out-when-productizing-cratedb:
+
+###############################################
+ Topics to watch out when productizing CrateDB
+###############################################
```
amotl, line 5: That's a bit of an outlier. Can you try to find a more concise headline and link label?

image 
image

amotl: It can be made different on the navigation vs. on the page headline, when needed. In this case, I think both spots can benefit.

amotl: Is "productizing" meant in the same sense like [Going into production](https://cratedb-guide--182.org.readthedocs.build/admin/going-into-production.html), but focusing on performance details instead of operational ones?

amotl: What about a possible shorter "Optimize for production" in this case, or, alternatively, a more general "Production notes"?

docs/performance/productizing.rst
```diff
@@ -0,0 +1,116 @@
...
+
+.. _section-joins:
+
+*******
+ JOINs
+*******
+
+CrateDB is a lot better at JOINs than many of our competitors and is getting
+better at every release, but JOINs in distributed databases are tricky to
+optimize, so in many cases queries involving JOINs may need a bit of tweaking.
```
amotl, line 108: I get the message, and I love it. However, I think the sound is a bit sloppy, specifically the bragging fragment

> is a lot better at JOINs than many of our competitors

even if it's true. Maybe we can find a better wording? /cc @kneth, @simonprickett

amotl: Maybe be more specific, possibly even to the extent of doing name dropping, like...

> CrateDB's engine to coordinate and execute distributed JOINs is already more efficient
> than many competing implementations {by other vendors | like ArangoDB, Cassandra, CockroachDB, MongoDB}, building upon years of maturing since its introduction with CrateDB 1.0.
> Because JOIN operations in distributed databases are tricky to optimize, 
> they need a bit of query tweaking, based on what the query planner says.

So, I've added another variant of bragging about CrateDB's implementation ("building upon years of maturing since its introduction with CrateDB 1.0"), and also another variant of sloppyness ("based on what the query planner says"). I am sure you will make the best out of it?

docs/performance/optimization.rst
```diff
@@ -0,0 +1,595 @@
...
+   FROM device_data
+   WHERE reading_time BETWEEN '2024-01-01' AND '2025-01-01'
+   ORDER BY reading_time DESC
+   LIMIT 10;
+
+use:
+
+.. code:: sql
+
+   SELECT reading_time,reading_value
```
hammerhead, line 204:
```suggestion
   SELECT reading_time, reading_value
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,595 @@
...
+   FROM device_data
+   WHERE reading_time BETWEEN '2024-12-20' AND '2025-01-01'
```
hammerhead, line 206: From what I see, the time range is the only difference. The section's title is "Only sort data when needed", it reads more like a "don't select more data than you need" example. The example could just omit `ORDER BY` entirely, as basically the last sentence before the example suggests?

docs/performance/optimization.rst
```diff
@@ -0,0 +1,595 @@
...
+transformations are only applied after the database has already worked out the
+final result set to be sent back to the client.
+
+So instead of:
+
+.. code:: sql
+
+   WITH mydata AS (
+     SELECT
+           date_format(device_data.reading_time) AS formatted_reading_time,
```
hammerhead, line 241:
```suggestion
           DATE_FORMAT(device_data.reading_time) AS formatted_reading_time,
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,595 @@
...
+However, CrateDB can do a lot better than this if we add an additional condition
+like this:
+
+.. code:: sql
+
+   SELECT *
+   FROM (
+      SELECT unnest(my_array_of_objects) obj
+      FROM my_table
+      WHERE 1 = ANY (my_array_of_objects['field1'])
```
hammerhead, line 149:
```suggestion
      SELECT UNNEST(my_array_of_objects) obj
      FROM my_table
      WHERE 1 = ANY(my_array_of_objects['field1'])
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,595 @@
...
+So instead of:
+
+.. code:: sql
+
+   WITH mydata AS (
+     SELECT
+           date_format(device_data.reading_time) AS formatted_reading_time,
+           device_data.reading_value
+     FROM device_data
+     )
```
hammerhead, line 244:
```suggestion
)
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,595 @@
...
+   SELECT *
+   FROM mydata
+   WHERE formatted_reading_time LIKE '2025%';
+
+use:
+
+.. code:: sql
+
+   SELECT
+     date_format(device_data.reading_time) AS formatted_reading_time,
```
hammerhead, line 254:
```suggestion
     DATE_FORMAT(device_data.reading_time) AS formatted_reading_time,
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,595 @@
...
+Replace CASE in expressions used for filtering, JOINs, grouping, etc
+====================================================================
+
+It is not always obvious to the optimizer what we may be trying to do with a
+``CASE`` expression (see for instance `Shortcut CASE evaluation Issue 16022`_).
+
+If you are using CASE expression for “formatting” see the previous point about
+formatting output as late as possible,
+
+but if you are using a CASE expression as part of a filter of other operation
```
hammerhead, line 270:
```suggestion
formatting output as late as possible,
but if you are using a CASE expression as part of a filter of other operation
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,595 @@
...
+
+(the exact replacement expressions of course depend on the semantics of each
+case)
+
+.. _groups-instead-distinct:
+
+Use groupings instead of DISTINCT
+=================================
+
+(Reference: `Issue 13818`_)
```
hammerhead, line 322:
```suggestion
(Reference: `Issue 13818`_)

Instead of
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,595 @@
...
+use
+
+.. code:: sql
+
+   SELECT COUNT(a)
+   FROM (
+           SELECT a
+           FROM t
+           GROUP BY a
+           ) tmp;
```
hammerhead, line 349:
```suggestion
   ) tmp;
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,595 @@
...
+
+If you need to perform lots of UPDATEs or expensive INSERTs from SELECT, instead
+of doing them all in one go, adopt a batch approach where the operations are
+done on groups of records each time.
+
+So for instance instead of doing:
+
+.. code:: sql
+
+   UPDATE mytable SET field1=field1+1;
```
hammerhead, line 401:
```suggestion
   UPDATE mytable SET field1 = field1 + 1;
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,595 @@
...
+
+do
+
+.. code:: shell
+
+   for id in {1..100}; do
+           crash -c "UPDATE mytable SET field1=field1+1 WHERE customer_id=$id;"
```
hammerhead, line 408:
```suggestion
           crash -c "UPDATE mytable SET field1 = field1 + 1 WHERE customer_id = $id;"
```

docs/performance/productizing.rst
```diff
@@ -0,0 +1,116 @@
...
+ Topics to watch out when productizing CrateDB
+###############################################
+
+.. _mindful-of-memory:
+
+*******************************
+ Be mindful of memory capacity
+*******************************
+
+In CrateDB operations requiring a working set like groupings, aggregations, and
```
hammerhead, line 13:
```suggestion
In CrateDB, operations requiring a working set like groupings, aggregations, and
```

docs/performance/productizing.rst
```diff
@@ -0,0 +1,116 @@
...
+sorting are performed fully in memory without spilling over to disk.
+
+Sometimes you may have a query that leads to a sub-optimal execution plan
+requiring lots of memory. If you are coming to CrateDB from other database
+systems, your experience may be that these queries will proceed to run taking
+longer than required and impacting other workloads in the meanwhile, sometimes
```
hammerhead, line 19:
```suggestion
longer than required and impacting other workloads in the meanwhile. Sometimes
```
To make sentences a bit shorter.

docs/performance/productizing.rst
```diff
@@ -0,0 +1,116 @@
...
+Please take a look at :ref:`Query Optimization 101 <performance-optimization>` for strategies to optimize your
+queries when you encounter this situation.
+
+.. _reading-lots-of-records:
+
+*************************
+ Reading lots of records
+*************************
+
+When the http endpoint is used CrateDB will prepare the entire response in
```
hammerhead, line 40:
```suggestion
When the HTTP endpoint is used CrateDB will prepare the entire response in
```

docs/performance/productizing.rst
```diff
@@ -0,0 +1,116 @@
...
+memory before sending it to the client.
+
+When the PostgreSQL protocol is used CrateDB attempts to stream the results but
+in many cases it still needs to bring all rows to the query handler node first.
+
+So we should always limit how many rows we request at a time, see `Fetching
+large result sets from CrateDB`_
```
hammerhead, line 47:
```suggestion
So we should always limit how many rows we request at a time, see `Fetching
large result sets from CrateDB`_.
```

docs/performance/productizing.rst
```diff
@@ -0,0 +1,116 @@
...
+call shards.
+
+If we do not specify how many shards we want for a table/partition CrateDB will
+derive a default from the number of CPUs.
+
+CrateDB also has replicas of data and this results in additional shards in the
+cluster.
+
+Having too many or too few shards has performance implications, so it is very
+important to get familiar with the :ref:`Sharding Performance Guide <sharding_guide>`
```
hammerhead, line 65:
```suggestion
important to get familiar with the :ref:`Sharding Performance Guide <sharding_guide>`.
```

docs/performance/productizing.rst
```diff
@@ -0,0 +1,116 @@
...
+
+******************
+ Number of shards
+******************
+
+In CrateDB data in tables and partitions is distributed in storage units that we
+call shards.
+
+If we do not specify how many shards we want for a table/partition CrateDB will
+derive a default from the number of CPUs.
```
hammerhead, line 59:
```suggestion
If we do not specify how many shards we want for a table/partition CrateDB will
derive a default from the number of nodes.
```

https://cratedb.com/docs/crate/reference/en/latest/sql/statements/create-table.html#clustered:
> num_shards = max(4, num_data_nodes * 2)

docs/performance/productizing.rst
```diff
@@ -0,0 +1,116 @@
...
+ Amount of (indexed) columns
+*****************************
+
+CrateDB indexes everything by default, providing excellent query performance
+without requiring manual indexing. However, excessive indexing can impact
+storage, write speed, and resource utilization.
+
+-  All columns are automatically indexed when inserted.
+-  CrateDB optimizes indexing using Lucene-based columnar storage.
+-  A soft limit of 1,000 indexed columns (or JSON fields) per table exists.
```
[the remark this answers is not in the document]
hlcianfagna: You are right, this does not seem to apply to first level columns:
```sql
CREATE TABLE t1 (a int);
    ALTER TABLE t1 SET ("mapping.total_fields.limit"=3);
    ALTER TABLE t1 ADD column b int;
    ALTER TABLE t1 ADD column c int;
    ALTER TABLE t1 ADD column d int INDEX OFF STORAGE WITH (columnstore = false);
--> SQLParseException[Limit of total columns [3] in table [doc.t1] exceeded]
```

This paragraph was originally from @wierdvanderhaar , we should edit it to make it about `OBJECT`s.

docs/performance/optimization.rst
```diff
@@ -0,0 +1,598 @@
...
+Indexing in CrateDB is optimized to support filtering and aggregations without
+requiring expensive defragmentation operations, but it is not optimized for
+sorting​.
+
+Maintaining a sorted index would slow down ingestion, that is why​ other
+analytical database systems like Cassandra and Redshift make similar
+trade-offs​.
+
+This means that when an ``ORDER BY`` operation is requested, the whole dataset
+needs to be loaded into the main memory on the relevant cluster node to be sorted.
```
seut, line 182: This isn't correct as CrateDB will push down any ``ORDER BY` as much as possible, in best case using an optimized lucene collector. Additional, intermediated node based ordering is applied and results from nodes are merged by a sorted merge on the handler node.

docs/performance/optimization.rst
```diff
@@ -0,0 +1,598 @@
...
+datasets into manageable steps.
+
+.. _batch-operations:
+
+Batch operations
+================
+
+If you need to perform lots of UPDATEs or expensive INSERTs from SELECT, instead
+of doing them all in one go, adopt a batch approach where the operations are
+done on groups of records each time.
```
seut, line 398: What is the reason to do this? CrateDB uses an internal batching and throttling logic to run this in the most efficient way. Is there an issue with this?

docs/performance/optimization.rst
```diff
@@ -0,0 +1,598 @@
...
+This section discusses expressions that improve filter efficiency and handling
+of specific data Structures.
+
+************************************
+ Filter and Expression Optimization
+************************************
+
+.. _positive-filters:
+
+Prefer positive filter expressions to negative filter expressions
```
seut, line 509: Is this really called like that? Isn't this more popular as "Avoid expression negation"`

docs/performance/productizing.rst
```diff
@@ -0,0 +1,116 @@
...
+systems, your experience may be that these queries will proceed to run taking
+longer than required and impacting other workloads in the meanwhile. Sometimes
+this effect may be obvious if a query takes a lot of resources and runs for a
+long time, other times it may go unnoticed if a query that could complete in say
+100 milliseconds takes one hundred times longer, 10 seconds, but the users put
+up with it without reporting to you.
+
+In CrateDB if a query would require more heap memory than the interested nodes
+have available the query will fail with a particular type of error message that
+we call a `CircuitBreakerException`, this is a fail-fast approach, we can see
```
seut, line 27: Why is this a fail-fast approach? It fails once the memory is occupied at runtime, could happen very late while processing queries. The main point here is: preventing the node from dying with an OOM.

docs/performance/productizing.rst
```diff
@@ -0,0 +1,116 @@
...
+
+*************************
+ Reading lots of records
+*************************
+
+When the HTTP endpoint is used CrateDB will prepare the entire response in
+memory before sending it to the client.
+
+When the PostgreSQL protocol is used CrateDB attempts to stream the results but
+in many cases it still needs to bring all rows to the query handler node first.
```
seut, line 44: Why *attempts*? With PG, it will always stream data. Also inter-node data is paged, depending on the query (e.g the handler node does not have to apply any pipeline breaker like an final in-memory ordering), the data is also paged directly to the client. Thus, the handler node does not need to hold the complete result set in-memory in most cases.

docs/performance/productizing.rst
```diff
@@ -0,0 +1,116 @@
...
+
+******************
+ Number of shards
+******************
+
+In CrateDB data in tables and partitions is distributed in storage units that we
+call shards.
+
+If we do not specify how many shards we want for a table/partition CrateDB will
+derive a default from the number of nodes.
```
seut, line 59: The default is **4**, it will **not** derive it from the number of nodes.

hlcianfagna: https://cratedb.com/docs/crate/reference/en/latest/sql/statements/create-table.html#clustered
says:

> If not provided, the number of shards is calculated based on the number of currently active data nodes

docs/performance/productizing.rst
hlcianfagna: > It fails once the memory is occupied at runtime

In some cases yes, but in other cases I believe it does an estimate and it returns with the `CircuitBreakerException` right away based on the estimate, maybe you could check this?

docs/performance/optimization.rst
```diff
@@ -0,0 +1,598 @@
...
+analytical database systems like Cassandra and Redshift make similar
+trade-offs​.
+
+This means that when an ``ORDER BY`` operation is requested, the whole dataset
+needs to be loaded into the main memory on the relevant cluster node to be sorted.
+That is why it is important to not request ``ORDER BY`` operations when not actually
+needed, and most importantly, not on tables of large cardinalities without aggregating
+records beforehand. On the other hand, of course it is no problem to sort a few
+thousand rows in the final stage of a ``SELECT`` operation, but we need to avoid
+requesting sort operations over millions of rows.
```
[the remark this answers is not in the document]
hlcianfagna: Is the case of the straightforward column references actually optimized in a way that we would not want to have this paragraph? the scenario we have in mind here is the `ORDER BY ts DESC LIMIT 100` from https://github.com/crate/crate/issues/16782

docs/performance/productizing.rst
seut: Estimates can only be done once the data is processed. When using e.g. literals, the value is available already and the estimate can be done immediately.

seut: In what cases did you experience "immediate" CBE's?

docs/performance/productizing.rst
seut: Oh right, my fault! 🤦  I should probably develop more often with > 4 nodes ;)

docs/performance/optimization.rst
hlcianfagna: Neither "negative filter expressions" nor "expression negation" seems particularly popular.
I changed the title to "Avoid expression negation in filters" now.
The underlying and more comprehensive technical concept is [sargability](https://en.wikipedia.org/wiki/Sargable), but I think we probably just want to keep it simple.

docs/performance/optimization.rst
```diff
@@ -0,0 +1,595 @@
+.. _performance-optimization:
+
+########################
+ Query Optimization 101
+########################
```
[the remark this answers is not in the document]
hlcianfagna: I added a brief introduction to both pages now, please review if you have a minute.

docs/performance/productizing.rst
hlcianfagna: I changed it to the shorter "Key design considerations for scaling" now, but it seems it is still too long?

docs/performance/productizing.rst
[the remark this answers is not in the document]
hlcianfagna: I updated this section now, please review if you have a chance.

docs/performance/optimization.rst
```diff
@@ -0,0 +1,609 @@
...
+Propagate LIMIT clauses when applicable
+=======================================
+
+Similarly to the above, we may have for instance a ``LIMIT 10`` at the end of
+the query and to get there it may have been sufficient to only pull 10 records
+(or some other number of records) at an earlier stage from some given table. If
+that is the case duplicate or move (depending on the specific query) the
+``LIMIT`` clause to the relevant place.
+
+In some cases, we may not know how many rows we need in the intermediate working
```
kneth, line 82: It is a long sentence. Breaking it up could ease reading it.

> In some cases, we may not know how many rows we need in the intermediate working sets but we know that there will be 10 records on the last day. Doing filtering early will help the optimizer and can protect the database from accidentally processing years of data. By not filtering early, the load on your cluster will increase tremendously.

docs/performance/optimization.rst
```diff
@@ -0,0 +1,609 @@
...
+requiring expensive defragmentation operations, but it is not optimized for
+sorting​.
+
+Maintaining a sorted index would slow down ingestion, that is why​ other
+analytical database systems like Cassandra and Redshift make similar
+trade-offs​.
+
+This means that when an ``ORDER BY`` operation is requested, the whole dataset
+needs to be loaded into the main memory on the relevant cluster node to be
+sorted. That is why it is important to not request ``ORDER BY`` operations when
```
kneth, line 190:
```suggestion
sorted. For this reason, it is important to not request ``ORDER BY`` operations when
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,609 @@
...
+
+CrateDB's query optimizer attempts to determine the most efficient way to
+execute a given query by considering the possible query plans. Based on the
+query scenario/situation, it is always aiming to use existing indexes on the
+original data for maximum efficiency.
+
+However, there is always a chance that some particular clause in the query
+expression prevents the optimizer from selecting an optimal plan, ending up
+applying the transformation on thousands or millions of records that later would
+be discarded anyway. So, whenever it makes sense, we want to make sure these
```
kneth, line 242:
```suggestion
be discarded anyway. So, whenever it makes sense, we want to ensure these
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,609 @@
...
+performance.
+
+.. _consider-generated-columns:
+
+Consider generated columns
+==========================
+
+If you frequently find yourself extracting information from fields and then
+using this extracted data on filters or aggregations, it can be good to consider
+doing this operation on ingestion with a `generated column`_, this way the value
```
kneth, line 480:
```suggestion
doing this operation on ingestion with a `generated column`_ . In this way the value
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,609 @@
...
+When using user-defined functions (UDFs), two important details relevant for
+performance aspects need to be considered.
+
+#. Once values are processed by an UDF, the database engine will load results
+   into memory, and will not be able to leverage indexes on the underlying
+   fields any longer. In this spirit, please apply the relevant general
+   considerations about delaying formatting as much as possible.
+
+#. UDFs run on a JavaScript virtual machine on a single thread, so they can have
+   an impact on performance, because relevant operations can not be
```
kneth, line 503:
```suggestion
   an impact on performance as relevant operations can not be
```

docs/performance/optimization.rst
```diff
@@ -0,0 +1,609 @@
...
+   WHERE customerid > 3 AND status = 'active';
+
+.. _use-null-or-empty:
+
+Use the special null_or_empty function with OBJECTs and ARRAYs when relevant
+============================================================================
+
+CrateDB has a special scalar function called null_or_empty_ , using this in
+filter conditions against OBJECTs and ARRAYs is much faster than using an ``IS
+NULL`` clause, if accepting empty objects and arrays is acceptable.
```
kneth, line 556:
```suggestion
NULL`` clause, if allowing empty objects and arrays is acceptable.
```

docs/performance/productizing.rst
```diff
@@ -0,0 +1,126 @@
...
+Sometimes you may have a query that leads to a sub-optimal execution plan
+requiring lots of memory. If you are coming to CrateDB from other database
+systems, your experience may be that these queries will proceed to run taking
+longer than required and impacting other workloads in the meanwhile. Sometimes
+this effect may be obvious if a query takes a lot of resources and runs for a
+long time, other times it may go unnoticed if a query that could complete in say
+100 milliseconds takes one hundred times longer, 10 seconds, but the users put
+up with it without reporting to you.
+
+In CrateDB if a query would require more heap memory than the interested nodes
```
kneth, line 29:
```suggestion
If a query would require more heap memory than the interested nodes
```

docs/performance/productizing.rst
```diff
@@ -0,0 +1,126 @@
...
+have available the query will fail with a particular type of error message that
+we call a ``CircuitBreakerException``, this is a fail-fast approach, we can see
```
kneth, line 31:
```suggestion
we call a ``CircuitBreakerException``. This is a fail-fast approach as we quickly see
```

docs/performance/productizing.rst
```diff
@@ -0,0 +1,126 @@
...
+quickly that there is an issue and optimize the query to get the best
```
kneth, line 32:
```suggestion that there is a performance issue and optimize the query to get the best
```

kneth (approved):
Thank you. I learned more about CrateDB performance tuning 🙇🏻

docs/performance/productizing.rst
amotl: It's still a bit long, yes. But we can also change it after an initial merge.

docs/performance/optimization.rst
amotl: Maybe it should better read/say "[...] instead of invoking individual queries, please consider adopting a batch approach [...]"?

kneth (approved):

docs/performance/optimization.rst
```diff
@@ -0,0 +1,610 @@
...
+So instead of:
+
+.. code:: sql
+
+   WITH mydata AS (
+     SELECT
+           DATE_FORMAT(device_data.reading_time) AS formatted_reading_time,
+           device_data.reading_value
+     FROM device_data
+)
```
amotl, line 255: This will fix the build error.
```suggestion
   )
```

docs/performance/productizing.rst
amotl, line 1: Even if it's one more nitpick, I think "productizing", as used in the filename, is the wrong word.

However, I don't want to bother you too much about it: As long as you don't quickly spread the word about the new resource URL, I can easily run a follow-up patch for you to review that adjusts a bit more things on the "naming things" details if you don't mind.

amotl: What about calling the file just `scaling.rst`, and use an imperative but concise "Design for scaling" as a title? Or would "Design for scale" be better/correct?

amotl:
@karynzv: Please squash all the commits before merging.

amotl:
Hi. @karynzv asked for support here, because CI is still failing. I will take over the branch, squash commits, and merge it when possible.

amotl:
I probably just destroyed the commit history 💥. Apologies. Can I humbly ask you to `--force` push from your workstation once again, @karynzv?

PR #185, opened 2025-03-21
Autogenerated sequences for PRIMARY KEY values

hlcianfagna:
## About

Adding a new section about:

- Autogenerated sequences and PRIMARY KEY values in CrateDB

## Preview

- https://cratedb-guide--185.org.readthedocs.build/performance/inserts/sequences.html


docs/performance/inserts/sequences.rst
```diff
@@ -0,0 +1,179 @@
...
+
+**Cons:** No order guarantee, not as human-friendly as numbers
+
+************************
+ Use UUIDv7 identifiers
+************************
+
+`Version 7 UUIDs`_ are a relatively new kind of UUIDs which feature a
+time-ordered value. We can use these in CrateDB with an UDF_ with the code from
+https://github.com/nalgeon/uuidv7/pull/37 (update once merged)
```
amotl, line 57: Excellent. It was merged one hour ago.

amotl (commented):
I didn't review the text in detail yet, but I love so much that there will be a dedicated page now.

Because the topic is not necessarily about performance matters, but more about functional details, do you see a chance to slot it in differently, or are you open to any proposals of ours?

docs/performance/inserts/sequences.rst
```diff
@@ -0,0 +1,179 @@
...
+   );
+
+The Python code below reads the last value used from the sequences table, and
+then attempts an `optimistic UPDATE`_ with a ``RETURNING`` clause, if a contending
+process already consumed the identity nothing will be returned so our process
+will retry until a value is returned, then it uses that value as the new ID for
+the record we are inserting into the ``mytable`` table.
+
+.. code:: python
+
```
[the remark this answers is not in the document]
amotl: Then, just invoke the program per `uv run <program.py>`.

docs/performance/inserts/sequences.rst
```diff
@@ -0,0 +1,179 @@
...
+.. code:: psql
+
+   BIGINT DEFAULT now() PRIMARY KEY
+
+**Pros:** Always increasing number - ideal if we need to timestamp records
+creation anyway
+
+**Cons:** gaps between the numbers, not suitable if we may have more than one
+record on the same millisecond
```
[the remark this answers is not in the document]
amotl: In Sphinx/docutils jargon, this language construct is called [Field Lists](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#field-lists). It is traditionally used to annotate function arguments in inline documentation.
```
:field name: field body
```

docs/performance/inserts/sequences.rst
```diff
@@ -0,0 +1,179 @@
...
+This option involves declaring a column as follows:
+
+.. code:: psql
+
+   TEXT DEFAULT gen_random_text_uuid() PRIMARY KEY
+
+**Pros:** Globally unique, no risk of conflicts if merging things from different
+tables/environments
+
+**Cons:** No order guarantee, not as human-friendly as numbers
```
amotl, line 49:
```suggestion
**Cons:** No order guarantee. Not as human-friendly as numbers. String format may
not be applicable to cover all scenarios.
```

docs/performance/inserts/sequences.rst
amotl, line 34:
```suggestion
**Cons:** Gaps between sequence numbers. Limited to millisecond granularity.
```

docs/performance/inserts/sequences.rst
```diff
@@ -0,0 +1,179 @@
...
+https://cratedb.com/blog/replicating-data-from-other-databases-to-cratedb-with-debezium-and-kafka
+for an example.
+
+*********************
+ Implement sequences
+*********************
+
+This approach involves a table to keep the latest values that have been consumed
+and client side code to keep it up-to-date in a way that guarantees unique
+values even when many ingestion processes run in parallel.
```
amotl, line 82:
```suggestion
Implementing sequences using a manual approach involves:
- A bookkeeping table to maintain the latest values that have been consumed.
- Client-side code to keep it up-to-date in a way that guarantees unique
  values even when many ingestion processes run in parallel.
```

docs/performance/inserts/sequences.rst
```diff
@@ -0,0 +1,179 @@
...
+the record we are inserting into the ``mytable`` table.
+
+.. code:: python
+
+   import records
+
+   db = records.Database("crate://")
+   sequence_name = "mysequence"
+
+   while True:
```
amotl, line 130: That reads dangerous, potentially implementing a busy loop, but I guess you know what you are doing?
Still, shall we possibly add a timeout on this identifier-polling operation, so the loop may not accidentally spin forever?

docs/performance/inserts/sequences.rst
```diff
@@ -0,0 +1,179 @@
...
+ Use IDs from an external system
+*********************************
+
+In cases where data is flowing to CrateDB from another system, we can declare
+the column on the CrateDB side without any defaults and just ingest the
+auto-incrementing values coming from the source database.
+
+See
+https://cratedb.com/blog/replicating-data-from-other-databases-to-cratedb-with-debezium-and-kafka
+for an example.
```
amotl, line 74:
```suggestion
See [Replicating data from other databases to CrateDB with Debezium and Kafka] for an example.

[Replicating data from other databases to CrateDB with Debezium and Kafka]: https://cratedb.com/blog/replicating-data-from-other-databases-to-cratedb-with-debezium-and-kafka
```

docs/performance/inserts/sequences.rst
```diff
@@ -0,0 +1,179 @@
...
+
+************************
+ Use UUIDv7 identifiers
+************************
+
+`Version 7 UUIDs`_ are a relatively new kind of UUIDs which feature a
+time-ordered value. We can use these in CrateDB with an UDF_ with the code from
+https://github.com/nalgeon/uuidv7/pull/37 (update once merged)
+
+**Pros:** Same as gen_random_text_uuid above but almost sequential
```
amotl, line 59:
```suggestion
**Pros:** Same as `gen_random_text_uuid` above but almost sequential
```

docs/performance/inserts/sequences.rst
amotl:
```suggestion
time-ordered value. We can use it in CrateDB with an UDF_ and the code from [uuidv7.cratedb].

[uuidv7.cratedb]: https://github.com/nalgeon/uuidv7/blob/main/src/uuidv7.cratedb
```

docs/performance/inserts/sequences.rst
```diff
@@ -0,0 +1,179 @@
...
+**Cons:** not as human-friendly as numbers and slight performance impact from
+UDF use
+
+*********************************
+ Use IDs from an external system
+*********************************
+
+In cases where data is flowing to CrateDB from another system, we can declare
+the column on the CrateDB side without any defaults and just ingest the
+auto-incrementing values coming from the source database.
```
amotl, line 70:
```suggestion
In cases where data is imported into CrateDB from external systems that employ
identifier governance, CrateDB does not need to generate any identifier values at
all. In this case, primary key information is inherited from the source system.
```

docs/performance/inserts/sequences.rst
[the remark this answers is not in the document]
amotl: It might not be applicable in this specific case, if it adds too much header noise.
However, you may find it helpful in other situations, so please read it merely FYI.

amotl (commented):
Thank you again. 👍

- I've added a few suggestions, many of them just formatting nitpicks. Feel free to take or reject each item at your disposal. 🙏
- I've reviewed most of the prose, i.e. no SQL, and also omitted the preamble up until now, focusing on the main body of the document.

hlcianfagna:
> Because the topic is not necessarily about performance matters, but more about functional details, do you see a chance to slot it in differently, or are you open to any proposals of ours?

I thought about it, but did not see any obvious place where it would fit naturally, I am happy to move it if you have any location in mind.

amotl:
>> Do you see a chance to slot it in differently?
>
> I thought about it, but did not see any obvious place where it would fit naturally.

I agree. There is no other proper place yet. Maybe we will repurpose the `performance` section in the future to accompany **both** performance and usability/convenience aspects?

Please go ahead slotting it in here, and let's have a quick conversation how to evolve this documentation section this or next week?

docs/performance/inserts/sequences.rst
```diff
@@ -0,0 +1,182 @@
...
+
+.. code:: psql
+
+   TEXT DEFAULT gen_random_text_uuid() PRIMARY KEY
+
+**Pros:** Globally unique, no risk of conflicts if merging things from different
+tables/environments
+
+**Cons:** No order guarantee. Not as human-friendly as numbers. String format may
+not be applicable to cover all scenarios.
```
seut, line 50: Additionally, I think it's worth adding that no range queries are possible, as UUIDs aren't sequential. This is one of the main benefits of algorithms like a [snowflake_id](https://en.wikipedia.org/wiki/Snowflake_ID)

docs/performance/inserts/sequences.rst
```diff
@@ -0,0 +1,182 @@
+.. _autogenerated_sequences_performance:
+
+###########################################################
+ Autogenerated sequences and PRIMARY KEY values in CrateDB
+###########################################################
+
+If you are coming to CrateDB from systems like PostgreSQL or MySQL one thing you
```
kneth, line 7: I have tried to rephrase the intro/motivation a bit:

As you begin working with CrateDB, you might be puzzled why CrateDB does not have a built-in, auto-incrementing "serial" data type as PostgreSQL or MySQL.

As a distributed database , designed to scale horizontally, CrateDB needs as many operations as possible to complete independently on each node without any coordination between nodes. 

Maintaining a global auto-increment value will require that a node would need to check with all other nodes before allocating a new value - and preventing all other nodes to allocate new values. This bottleneck will be hindering our ability to achieve extremely_fast_ingestion_speeds.

kneth (approved):
Thank you!!

docs/performance/inserts/sequences.rst
```diff
@@ -0,0 +1,179 @@
...
+           str(
+               db.query(
+                   update_query,
+                   new_value=new_value,
+                   sequence_name=sequence_name,
+                   seq_no=row._seq_no,
+                   primary_term=row._primary_term,
+               ).all()
+           )
+           != "[]"
```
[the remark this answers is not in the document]
hlcianfagna: I really looked for an alternative but could not find it. There would be plenty of options giving up on the records library but I like that it makes the rest of the code so simple.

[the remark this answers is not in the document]
amotl: Thanks. It certainly should not block the merge. Because I love that you also like the `records` library, I may check this spot on a subsequent iteration to possibly find a better way to compare for empty?

Why not just do it this way, if it's semantically correct that you want to break if the result is **not** empty?
```python
result = db.query(....)
if result:
    break
```

docs/performance/inserts/sequences.rst
```diff
@@ -0,0 +1,196 @@
...
+ Use IDs from an external system
+*********************************
+
+In cases where data is imported into CrateDB from external systems that employ
+identifier governance, CrateDB does not need to generate any identifier values
+and primary key values can be inserted as-is from the source system.
+
+See [Replicating data from other databases to CrateDB with Debezium and Kafka] for an example.
+
+[Replicating data from other databases to CrateDB with Debezium and Kafka]: https://cratedb.com/blog/replicating-data-from-other-databases-to-cratedb-with-debezium-and-kafka
```
amotl, line 83: That's Markdown. ;]

NB: We will convert this file and others to Markdown anyway, to provide better support for [llms.txt](https://github.com/crate/tech-content/issues/135), so it does not necessarily needs to be fixed.

amotl (approved):
Approving, thanks again! My suggestions for a later iteration are:

- Relocate the document into the future "Data Modeling" subsection, like planned with @kneth.
- Convert to Markdown.
- More sanding.

NB: Let me know if you need support about technicalities, for example wrestling with reStructuredText.

amotl:
@coderabbitai: In your comment above, you said:

> It may be helpful to explicitly state that UUIDs do not support efficient range queries, which is a key limitation compared to sequential numeric IDs. This is especially relevant for users considering query patterns.

However, here we are using the UUIDv7 variant, which indeed supports efficient range queries, because subsequently generated tokens are sequential like numeric IDs.

docs/performance/inserts/sequences.rst
```diff
@@ -0,0 +1,196 @@
...
+.. code:: psql
+
+   BIGINT DEFAULT now() PRIMARY KEY
+
+:Pros: Always increasing number - ideal if we need to timestamp records
+creation anyway
+
+
+:Cons: gaps between the numbers, not suitable if we may have more than one
+record on the same millisecond
```
amotl, line 36: This snippet causes the _WARNING: Field list ends without a blank line; unexpected unindent._

-- https://github.com/crate/cratedb-guide/actions/runs/14473917871/job/40594700368?pr=185#step:4:581

I think this might work better:
```suggestion
:Pros: Always increasing number - ideal if we need to timestamp records
       creation anyway
:Cons: gaps between the numbers, not suitable if we may have more than one
       record on the same millisecond
```

PR #199, opened 2025-05-08
Update index.rst

bmunkholm:
Remove pointless sentence - doesn't add any value.

PR #200, opened 2025-05-08
Fix broken links: Documentation about CrateDB Edge has been removed

amotl:
## About
What the title says.

## References
- https://github.com/crate/cloud-docs/pull/97


bmunkholm (approved):
Thanks again!

amotl:
> For my education, I wonder if we don't have any link checking across the various documentation projects?

We do. It is exactly the link checker which was tripping on this occasion, as observed on your other patch GH-199.

> And subsequent what the best practices are to ensure all links between doc projects are good?

The common practice is to fix forward, by submitting patches like this one.

PR #201, opened 2025-05-09
CI: Remove building docs on macOS

amotl:
## About
The CI workflows' intentions here are mostly to run the link checker, so it's not too much connected to covering operating systems.

By removing the checks for macOS, the process will relax pressure on downstream resources, aiming for making it more unlikely to hit rate limiting measures.

## Details
It's not just rate limiting, it's also **429 Too Many Requests**.
```
(integrate/langchain/index: line  141) broken    https://github.com/crate/cratedb-examples/blob/main/topic/machine-learning/llm-langchain/conversational_memory.ipynb -
429 Client Error: Too Many Requests for url: https://github.com/crate/cratedb-examples/blob/main/topic/machine-learning/llm-langchain/conversational_memory.ipynb
```
-- https://github.com/crate/cratedb-guide/actions/runs/14926973334/job/41933954698?pr=199#step:4:1689

## References
- https://github.com/crate/cratedb-guide/pull/199#issuecomment-2866067152

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -21,7 +21,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest]
 
     steps:
       - name: Acquire sources


.github/workflows/docs.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest]
```
amotl, line 24: Let's please propagate this change also to other projects which are doing it the same way, in order to reduce overall resource consumption.

bmunkholm (approved):
I was wondering wondering why we did this in the first place :-)

amotl:
> I was wondering wondering why we did this in the first place :-)

We just inherited this from previous generations of hackers. It might have made sense to integration-test things like 
[crate-docs](https://github.com/crate/crate-docs) also here on downstream projects while it was still in flux, and also it might have made sense in previous eras because Sphinx might not always have been stable for macOS. We need to consider here that the whole operation started in Sphinx 1.x times already, around a decade ago.

Essential ingredients of the documentation stack are certainly more stable across operation systems today, so it will not be a huge loss to remove this slot from the test matrix.

PR #202, opened 2025-05-09
cleanup of docs.yml

bmunkholm:
Slight cleanup of the build file for a better blueprint for similar repos.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,13 +15,8 @@ concurrency:
 
 jobs:
   documentation:
-
-    name: Build docs on ${{ matrix.os }}
-    runs-on: ${{ matrix.os }}
-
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -15,13 +15,8 @@ concurrency:
...
 jobs:
   documentation:
-
-    name: Build docs on ${{ matrix.os }}
-    runs-on: ${{ matrix.os }}
-
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
+    name: Build docs on ubuntu-latest
```
amotl, line 18: Might be good enough to optionally make it even shorter.

```suggestion
    name: Build docs
```

amotl (approved):

PR #203, opened 2025-05-13
Remove duplicate resource links in index.md

bmunkholm:

The "Resources" contained links to information that was already part of the page. So I removed those duplicate links away for simplicity. So now it's focused on "additional" resources not mentioned on the page already.

UPDATE:
I ended up removing them entirely as they are all both present in the left navbar or the top menu bar and in addition mentioned inside the page.


amotl (approved):
Thanks!

PR #213, opened 2025-06-11
Dedicated `CircuitBreaker` page

karynzv:
The goal is to add a page that has a complete guide on CB and most importantly the CB exception info, to provide users with the context on why that happens and what they should do about it.

Please, feel free to edit it and provide feedback so we can improve this and make it as accurate as possible.
## Checklist

 - [ ] Link to issue this PR refers to (if applicable): Fixes #???

### DIFF
--- a/docs/admin/circuit-breaker.rst
+++ b/docs/admin/circuit-breaker.rst
@@ -23,11 +23,14 @@ A query executes as an ordered set of operations. Before running each stage, Cra
 If the projected total would exceed the breaker limit, the system aborts the query and returns a ``CircuitBreakingException``. 
 This pre-emptive trip prevents the JVM's garbage collector from reaching an unrecoverable out-of-memory state.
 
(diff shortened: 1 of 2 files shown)


hlcianfagna (approved):

docs/admin/circuit-breaker.rst
```diff
@@ -0,0 +1,94 @@
...
+consult the `CircuitBreakers MXBean`_ section.
+
+For hosted deployments, see `Cloud monitoring`_ and for self-managed clusters, the `on-prem monitoring`_ guide. Both describe the complete path from cluster
+deployment to collecting metrics and displaying them on a Grafana dashboard.
+
+Exception Handling
+==================
+.. code-block:: console
+  
+  CircuitBreakingException: [query] Data too large, data for [mergeOnHandler: 1] would be [9021423616/8.4gb], which is larger than the limit of [9019431321/8.3gb]
```
hammerhead, line 46: I think this is an outdated version of the error message. It was changed recently to remove the double numbers (`[9021423616/8.4gb]`) and hints what to do were added directly the the error.

PR #214, opened 2025-06-17
Storage Usage: Add benchmark results from v5.10.2

surister:
## About
Update the page about "Performance » Storage Usage" regarding storage space improvements with CrateDB 5.10.

## Preview
https://cratedb-guide--214.org.readthedocs.build/performance/storage.html

## Checklist
 - [x] Link to issue this PR refers to (if applicable): Fixes https://github.com/crate/tech-content/issues/136

### DIFF
--- a/docs/performance/storage.md
+++ b/docs/performance/storage.md
@@ -1,15 +1,17 @@
 # Storage Usage
 
-CrateDB stores data in a row and column store, and automatically creates an index on top of that.
-On reads, the index will be leveraged: Depending on the query, the engine will use the most efficient store.
+CrateDB stores data using both row and columnar stores, with automatic
+indexing on every column, enabling efficient searches without requiring
+manual index creation. Indexes are leveraged for reads, and depending on
+the query, the engine uses the most efficient store.
 
 This is one of the many features that makes CrateDB very fast when reading
 and aggregating data, but it has an impact on storage size.
 
-We are going to use [Yellow taxi trip - January 2024](https://www.nyc.gov/site/tlc/about/tlc-trip-record-data.page) which has 2_964_624 rows
+We are going to
+use [Yellow taxi trip - January 2024](https://www.nyc.gov/site/tlc/about/tlc-trip-record-data.page)
+which has 2_964_624 rows.
 
-
-```
 | VendorID | tpep_pickup_datetime | tpep_dropoff_datetime | passenger_count | trip_distance | RatecodeID | store_and_fwd_flag | PULocationID | DOLocationID | payment_type | fare_amount | extra | mta_tax | tip_amount | tolls_amount | improvement_surcharge | total_amount | congestion_surcharge | Airport_fee |
 |----------|----------------------|-----------------------|-----------------|---------------|------------|--------------------|--------------|--------------|--------------|-------------|-------|---------|------------|--------------|-----------------------|--------------|----------------------|-------------|
 | 2        | 1704073016000        | 1704074392000         | 4               | 6.88          | 1          | "N"                | 170          | 231          | 1            | 32.4        | 1     | 0.5     | 7.48       | 0            | 1                     | 44.88        | 2.5                  | 0           |
@@ -17,18 +19,27 @@ We are going to use [Yellow taxi trip - January 2024](https://www.nyc.gov/site/t
 | 1        | 1704071126000        | 1704071510000         | 2               | 1             | 1          | "N"                | 140          | 141          | 1            | 7.9         | 3.5   | 0.5     | 2.55       | 0            | 1                     | 15.45        | 2.5                  | 0           |
 | 2        | 1704072696000        | 1704073070000         | 1               | 1.03          | 1          | "N"                | 262          | 75           | 1            | 8.6         | 1     | 0.5     | 2.72       | 0            | 1                     | 16.32        | 2.5                  | 0           |
 | 2        | 1704074134000        | 1704074399000         | 1               | 1.08          | 1          | "N"                | 249          | 68           | 1            | 7.2         | 1     | 0.5     | 2.44       | 0            | 1                     | 14.64        | 2.5                  | 0           |
-```
 
-Will take:
+The taxi dataset takes:
 
 - ~48MiB in Parquet (very optimized for storage)
 - ~342MiB in CSV
 - ~1.2GiB in JSON
 - ~510MiB in PostgreSQL 16.1 (Debian 16.1-1.pgdg120+1)
 - ~775MiB in CrateDB 5.9.3 (3 nodes, default settings)
+- ~431MiB in CrateDB 5.10.9 (3 nodes, default settings)
 
-At first sight, it might look that CrateDB storage takes more than PostgreSQL,
-but we need to dive deeper to really understand what is going on.
+We will dive deeper to really understand what is going on.
+
+:::{note}
+In version 5.10 storage usage was improved, some users report up to 70% of storage reduction,
+more reasonably a 30–50% can be expected for most data, the reduction percentage will vary
+depending on the use case.
+
+Read more at https://cratedb.com/blog/reducing-storage-costs-in-cratedb-v5.10
+
+This guide has been updated to reflect the v5.10 improvements.
+:::
 
 ## How storage in CrateDB works.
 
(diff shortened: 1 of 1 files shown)


docs/performance/storage.md
```diff
@@ -19,26 +19,34 @@ We are going to use [Yellow taxi trip - January 2024](https://www.nyc.gov/site/t
...
 - ~775MiB in CrateDB 5.9.3 (3 nodes, default settings)
+- ~431MiB in CrateDB 5.10.2 (3 nodes, default settings)
+
+We will dive deeper to really understand what is going on.
 
-At first sight, it might look that CrateDB storage takes more than PostgreSQL,
-but we need to dive deeper to really understand what is going on.
+> INFO: In version 5.10 storage usage was improved, some users report up to 70% of storage reduction,
+> more reasonably a 30–50% can be expected for most data, the reduction % of reduction will vary
+> depending on use cases.
```
amotl, line 34: FYI: In MyST Markdown, you may want to phrase the [admonition](https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html) like this:
```suggestion
:::{note}
In version 5.10 storage usage was improved, some users report up to 70% of storage reduction,
more reasonably a 30–50% can be expected for most data, the reduction % of reduction will vary
depending on use cases. 
:::
```

docs/performance/storage.md
```diff
@@ -47,30 +55,29 @@ CrateDB is a distributed database; nodes, shards, partitions and replicas are ti
...
 obtained. For `PostgreSQL` it was straightforward:
 
 ```sql
 SELECT pg_size_pretty(pg_total_relation_size('taxi_january'));
 ```
 
-For CrateDB when a table is created, sharding and replication has to be taken into account.
-When a table is created with default values, it gets partitioned in `max(4, num_data_nodes * 2)` shards.
+In CrateDB, when a table is created, sharding and replication have to be taken into account.
+When a table is created with default values, it gets partitioned using this formula: `max(4, num_data_nodes * 2)`.
```
amotl, line 66: Maybe elegantly refer to the sharding and scaling guideline docs at this spot?

- https://cratedb.com/docs/guide/performance/sharding.html
- https://cratedb.com/docs/guide/performance/scaling.html#number-of-shards

docs/performance/storage.md
```diff
@@ -80,34 +87,43 @@ FROM sys.shards
...
 
-This can be checked locally; querying `select table_name, path from sys.shards` shows the file path of the shard.
+| avg_mib_per_shard | total_mib |
+|-------------------|-----------|
+| 35                | 431       |
+
+The average size of a shard is ~35MiB, `35 * 12 = ~420`.
+
+You can check the actual size of a shard locally; querying `select table_name, path from sys.shards`
+shows the file path of the shards.
```
amotl, line 116:
```suggestion
You can check the actual size of a shard by inspecting the filesystem.
Use `select table_name, path from sys.shards` to show the file path of the individual shards.
```

docs/performance/storage.md
```diff
@@ -143,8 +160,8 @@ CREATE TABLE taxi
 )
 ```
 
-The index can only be disabled when the table is created, if the table already exists and it cannot
-be deleted it will have to be re-created.
+The index can only be disabled when the table is created, if the table already exists, and it can’t
+be deleted, it will have to be re-created.
```
amotl, line 164: Is it possible to delete an index afterwards? Doesn't the table _always_ need to be re-created?

docs/performance/storage.md
```diff
@@ -171,15 +188,24 @@ DROP TABLE "taxi_deleteme"
 
 > WARNING: Dropping the table deletes the data, make sure that the copy was done correctly.
 
-> INFO: Indexes cannot be re-added after table creation.
+> INFO: Indexes can’t be re-added after table creation.
```
amotl, line 191:
```suggestion
:::{attention}
Indexes can’t be re-added after table creation.
:::
```

docs/performance/storage.md
```diff
@@ -204,11 +230,36 @@ CREATE TABLE IF NOT EXISTS "doc"."taxi_nocolumnstore"(
...
+CrateDB v5.10.2
+
+| avg_mib_per_shard | total_mib |
+|-------------------|-----------|
+| 39                | 471       |
+
+Data was reduced `~13.11%` in v5.9 similar to `no_index`.
+Data was increased `~10%` in v.5.10
+
+> NOTE: In v5.10, we don’t recommend that you turn off column storage.
```
amotl, line 248:
```suggestion
:::{note}
In v5.10, we don’t recommend that you turn off column storage.
:::
```

docs/performance/storage.md
```diff
@@ -204,11 +230,36 @@ CREATE TABLE IF NOT EXISTS "doc"."taxi_nocolumnstore"(
...
+
+## Upgrade to v5.10 and Optimize table
+
+In v5.10 we introduced a new table storage format that can reduce storage up to 50% in comparison
+to previous versions.
+
+The easiest way to gain massive storage reduction is to update to the latest version available.
```
amotl, line 255: I would suggest to remove the additional line break here.
```suggestion
In v5.10 we introduced a new table storage format that can reduce storage up to 50% in comparison
to previous versions. The easiest way to gain massive storage reduction is to update to the latest version available.
```

docs/performance/storage.md
```diff
@@ -204,11 +230,36 @@ CREATE TABLE IF NOT EXISTS "doc"."taxi_nocolumnstore"(
...
+In v5.10 we introduced a new table storage format that can reduce storage up to 50% in comparison
+to previous versions.
+
+The easiest way to gain massive storage reduction is to update to the latest version available. 
+
+Additionally, as part of CrateDB's background operations, index segments are merged and optimized;
+you would typically see storage usage being automatically reduced as more data gets added to the
+table. Nonetheless, you can trigger these optimizations manually by running:
+
+`OPTIMIZE table <table_name> WITH (max_num_segments=1)`
```
amotl, line 261: Just a little more uppercasing.
```suggestion
`OPTIMIZE TABLE <table_name> WITH (max_num_segments=1)`
```

docs/performance/storage.md
```diff
@@ -204,11 +230,36 @@ CREATE TABLE IF NOT EXISTS "doc"."taxi_nocolumnstore"(
...
+| avg_mib_per_shard | total_mib |
+|-------------------|-----------|
+| 39                | 471       |
+
+Data was reduced `~13.11%` in v5.9 similar to `no_index`.
+Data was increased `~10%` in v.5.10
+
+> NOTE: In v5.10, we don’t recommend that you turn off column storage.
+
+## Upgrade to v5.10 and Optimize table
```
amotl, line 250: Maybe?
```suggestion
## Upgrade to v5.10 and `OPTIMIZE TABLE`
```

docs/performance/storage.md
```diff
@@ -204,11 +230,36 @@ CREATE TABLE IF NOT EXISTS "doc"."taxi_nocolumnstore"(
...
+
+In v5.10 we introduced a new table storage format that can reduce storage up to 50% in comparison
```
amotl, line 252:
```suggestion
CrateDB v5.10 introduced a new table storage format that can reduce storage up to 50% in comparison
```

amotl (commented):
Thanks a stack for the update, this outcome looks very promising.

I've added a few suggestions about syntax and friends, and would like to give @hlcianfagna the pleasure of final approval, as he knows the topic matter and the English language better than me.

docs/performance/storage.md
surister: Well its bad wording, they cannot be added after a table is created without them; @hlcianfagna wording is better.

docs/performance/storage.md
surister: Ah, I see your point, will amend. `and it can’t
be deleted` goes out

docs/performance/storage.md
surister: You were right! It didn't change storage usage until I merged segments.

docs/performance/storage.md
```diff
@@ -204,11 +241,41 @@ CREATE TABLE IF NOT EXISTS "doc"."taxi_nocolumnstore"(
...
+
+| avg_mib_per_shard | total_mib |
+|-------------------|-----------|
+| 39                | 471       |
+
+Data was reduced `~13.11%` in v5.9 similar to `no_index`
+Data was increased `~10%` in v5.10
+
+:::{note}
+In v5.10, we don’t recommend that you turn off column storage.
```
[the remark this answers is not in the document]
surister: Yes, will do. I will mostly quote Alan:

Alan: _columnstore=off thing I would expect to not work any more, as if there's no columnstore then we need to store the data somewhere else, and columnstores will be more efficient in terms of disk usage as they are designed specifically for each data type.  Previously it was in the source and therefore duplicated, but now there is no source, so there's pretty much no advantage at all to disabling the column store._
Me: _That's why columnstore=off takes more space :)?_
Alan: _Yep. ..._

docs/performance/storage.md
```diff
@@ -299,17 +391,18 @@ CREATE TABLE IF NOT EXISTS "doc"."taxi_nocolumnstore_noindex_bestcompresion" (
 
 CrateDB's default settings are optimized for performance.
 
-If some columns will never be used for aggregations or groupings, there will be no performance penalty.
-That might change in the future as your use case and data needs evolve,
-re-adding indexes or column store at later stages will need re-creating tables, 
-which might need some downtime, depending on the setup.
+If some columns are never used for aggregations or groupings, there will be no performance penalty.
```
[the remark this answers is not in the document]
surister: I will follow up this later.

amotl:
Yeah, link checker is failing. I will submit a separate patch later. Feel free to merge.

docs/performance/storage.md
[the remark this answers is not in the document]
amotl: Thanks for explaining that!

Cheers, @romseygeek.

amotl:
What about this remark / condition?

> This branch cannot be rebased due to conflicts

PR #217, opened 2025-06-26
Remove page-local TOC

bmunkholm:
Final round of removing page-local TOC in support of https://github.com/crate/tech-content/issues/140.
Also fixing minor wording suggestions from CodeRabbit.


amotl (approved):
Thanks!

PR #249, opened 2025-08-13
Naming things: Simplify directory layout at "Integrate" section

amotl:
## About

The flat-list directory layout at `docs/integrate` used prefixes for certain product names. In the spirit of better navigation (less is more), let's strip the prefixes for the sake of less confusion when exploring/navigating and authoring: We've elaborated below about our slightly negative experience.

## Preview

https://cratedb-guide--249.org.readthedocs.build/integrate/ (you don't "see" too much here, only when navigating)

## Rationale

Names for Apache and Amazon/AWS products have been prefixed before using `apache-` and `aws-`. However, while initially planned to be a good idea, it implicitly added an overlay subgroup to this flat list which skewed the core idea.

On the one hand intended, it turned out to be cumbersome to navigate, both when authoring, and when browsing / exploring the list on behalf of the directory tree, URL layout, or other interfaces.

So, let's strip all the prefixes, where the product name is feasible to be better used standalone.

## Thoughts

Other than the feeling of a smell, those specific thought conflicts and copy-editing experiences contributed to this change, sorted by increasing gravity:

- Kafka is not only an Apache product, but also a technology that is provided by different implementations, both unmanaged and managed. I'd like to elaborate on this on the relevant page in a quick statement for name-dropping reasons.

- `aws-dynamodb` and `aws-kinesis` had also been defined "wrong", because the real product names are "Amazon DynamoDB" and "Amazon Kinesis", while `aws-dms` and `aws-lambda` are correct (AWS DMS, AWS Lambda). That would have been another conflict zone.

- When doing actual work on the new tree yesterday, when thinking about navigating to the "Kafka" or "Kinesis" slot, needed one additional thought step to find them in `apache-kafka` or `aws-kinesis`. That yielded a slight bad taste, not just once.


docs/integrate/airflow/index.md
amotl, line 1: Sure that needs some rewrite rules.

```nginx
rewrite ^/docs/guide/integrate/apache-(.*)/(.*)$                     /docs/guide/integrate/$1/$2           redirect;
rewrite ^/docs/guide/integrate/aws-(.*)/(.*)$                        /docs/guide/integrate/$1/$2           redirect;
```

amotl: Added to the set of redirects for the current iteration.

- https://github.com/crate/infrastructure/commit/365782e83

docs/integrate/kafka/index.md
```diff
@@ -1,5 +1,5 @@
 (apache-kafka)=
```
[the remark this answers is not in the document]
amotl: 1e0a1a035f23b adjusts the reference labels accordingly. Let's do the proper spring cleaning right away.

amotl:
>> Assuming build error gets fixed :-)
>
> Assuming anything without any crew? ;]

That's more background about this particular detail. We need to remove a compatibility layer we have been carrying forward up to today due to ~~laziness~~ understaffing.

- https://github.com/crate/crate-docs-theme/issues/553#issuecomment-3188191772

PR #261, opened 2025-08-17
Integrate/Oracle: Add section with starter tutorial

amotl:
## About
There was no section about how to import data from Oracle Database yet.

## Details
The goal is to present concise walkthroughs without many bells and
whistles, which get to the point of getting you started quickly.

Use the canonical template to present another data nozzle based on
the CTK Ingestr I/O subsystem.

## Preview
https://cratedb-guide--261.org.readthedocs.build/integrate/oracle/

### DIFF
--- a/docs/integrate/oracle/index.md
+++ b/docs/integrate/oracle/index.md
@@ -10,7 +10,7 @@
 :::{rubric} About
 :::
 
-[Oracle Database] (Oracle DBMS, or simply as Oracle) is a proprietary multi-model
+[Oracle Database] (Oracle DBMS, or simply Oracle) is a proprietary multi-model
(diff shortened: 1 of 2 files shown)


gvenzl:
> > [...] it looks like it is good to be complex to test this.
> 
> We have been so happy to discover ready-made OCI images elsewhere ;], otherwise surfacing this guide would have been an unfeasible operation. Thank you so much, @gvenzl. 💯
> 
> -- https://hub.docker.com/r/gvenzl/oracle-free
> 
> OT: We see you are also [working on MCP](https://github.com/gvenzl/mcp). Cheers!

You are very welcome, @amotl! :)

Issue #243, opened 2025-08-11
Documentation feedback on /docs/integrate/telegraf/index.md


amotl:
### Documentation feedback

 
- Page title: Telegraf
- Page URL: https://cratedb.com/docs/guide/integrate/telegraf/index.html
- Page URL: https://cratedb.com/blog/use-cratedb-with-telegraf-an-agent-for-collecting-reporting-metrics
- Source: https://github.com/crate/cratedb-guide/blob/main/docs/integrate/telegraf/index.md

---

## Problem

Other than the resources listed above, we don't show much about how to integrate with Telegraf well, and also on the [tutorial](https://cratedb.com/blog/use-cratedb-with-telegraf-an-agent-for-collecting-reporting-metrics), the main gist is obfuscated in favor of a full tutorial that also includes Grafana.

## Solution

Full-stack / end-to-end examples are nice, but let's have a dedicated page as a another main stage for this purpose, where the focus is on the integration of CrateDB with Telegraf itself, in order to highlight the **CrateDB output plugin**.

- Docs: https://docs.influxdata.com/telegraf/v1/plugins/#output-cratedb
- README: https://github.com/influxdata/telegraf/blob/release-1.35/plugins/outputs/cratedb/README.md
- Patch: https://github.com/influxdata/telegraf/pull/3210

The core gist seems to be this configuration snippet:

```toml
# Configuration for CrateDB to send metrics to.
[[outputs.cratedb]]
  # A github.com/jackc/pgx/v4 connection string.
  # See https://pkg.go.dev/github.com/jackc/pgx/v4#ParseConfig
  url = "postgres://crate@localhost/doc?sslmode=disable"
  # Timeout for all CrateDB queries.
  timeout = "5s"
  # Name of the table to store metrics in.
  table = "metrics"
  # If true, and the metrics table does not exist, create it automatically.
  table_create = true
  # The character(s) to replace any '.' in an object key with
  key_separator = "_"
```

Issue #227, opened 2025-08-06
Improve general guidance


amotl:
## About

> The developer documentation has grown organically over a decade. Its corpus is large, and for a first-time user of CrateDB, it can be difficult to find the information needed.

Improve relevant details.

PR #267, opened 2025-08-24
Integrate/Telegraf: Add micro tutorial. Improve layout on index page.

amotl:
## About
Using Telegraf and CrateDB to collect and store system metrics, all based on using Docker for executing any command.

## Preview
- https://cratedb-guide--267.org.readthedocs.build/integrate/telegraf/
- https://cratedb-guide--267.org.readthedocs.build/integrate/telegraf/tutorial.html
- https://cratedb-guide--267.org.readthedocs.build/ingest/telemetry/

## References
- GH-227
- GH-243
- GH-247 (similar)

/cc @hammerhead

### DIFF
--- a/docs/integrate/telegraf/tutorial.md
+++ b/docs/integrate/telegraf/tutorial.md
@@ -68,10 +68,16 @@ The database connection URL is a `pgx/v4` connection string. Configure
 `table_create = true` to automatically let Telegraf create the metrics table
 if it doesn't exist.
 ::::{tab-set}
-:::{tab-item} Linux and macOS
+:::{tab-item} Linux
 ```shell
 sed -i 's!postgres://user:password@localhost/schema?sslmode=disable!postgres://crate@cratedb/doc?sslmode=disable!g' telegraf.conf
 sed -i 's!# table_create = false!table_create = true!' telegraf.conf
+```
+:::
+:::{tab-item} macOS and BSD
+```shell
+sed -i '' 's!postgres://user:password@localhost/schema?sslmode=disable!postgres://crate@cratedb/doc?sslmode=disable!g' telegraf.conf
+sed -i '' 's!# table_create = false!table_create = true!' telegraf.conf
 ```
 :::
 :::{tab-item} Windows PowerShell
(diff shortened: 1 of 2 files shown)


amotl:
> How about making it work against Cloud instead (or as well - by default)

Only a [few](https://cratedb.com/docs/guide/integrate/mongodb/cloud.html) [tutorials](https://cratedb.com/docs/guide/integrate/influxdb/cloud.html) from our pen made a start to consider cloud details the other day. While this was meant as a blueprint how to start with documenting relevant details, we observe very little user testing on those details in general, despite asking for support so often.

Because other patches also don't take a particular focus,

- GH-263

... I'd suggest to merge early, and do relevant improvements on behalf of future iterations. Contributions are very much welcome.

docs/integrate/telegraf/tutorial.md
```diff
@@ -0,0 +1,123 @@
...
+  config > telegraf.conf
+```
+
+The database connection URL is a `pgx/v4` connection string. Configure
+`table_create = true` to automatically let Telegraf create the metrics table
+if it doesn't exist.
+::::{tab-set}
+:::{tab-item} Linux and macOS
+```shell
+sed -i 's!postgres://user:password@localhost/schema?sslmode=disable!postgres://crate@cratedb/doc?sslmode=disable!g' telegraf.conf
```
kneth, line 73: BSD sed and GNU sed do not have the same semantics for option `-i`.  I suggest to use `-i ''` to cover both cases.

kneth (approved):
I have only a small change request

docs/integrate/telegraf/tutorial.md
amotl: @coderabbiai already suggested the [same](https://github.com/crate/cratedb-guide/pull/267#pullrequestreview-3149345175), thanks!

In `docs/integrate/telegraf/tutorial.md`, we recommend to...

> `71-84`: **Split Linux vs macOS for sed; BSD sed needs a different -i form.**
>
> The current “Linux and macOS” block will fail on macOS because BSD sed requires `-i ''`. Also, consider setting the table name explicitly to avoid relying on implicit defaults.

Issue #102, opened 2024-07-21
Consolidate Integration Tutorials I vs. II


amotl:
## About
On both pages listed below, we are referring to [Integration Tutorials I](https://cratedb.com/docs/guide/integrate/index.html#integrate) vs. [Integration Tutorials II](https://community.cratedb.com/t/overview-of-cratedb-integration-tutorials/1015). It needs to be consolidated.

## References
- https://cratedb.com/docs/guide/home/
- https://cratedb-guide--98.org.readthedocs.build/migrate/rockset/

amotl:
Hereby, we are making a start on this refactoring process.

- https://github.com/crate/cratedb-guide/pull/150

NB: We will share a few patches as we go, that can be used as blueprints to learn how the refactoring works. Because it's all Markdown, it works pretty straight-forward on the first iteration when copying content verbatim.

/cc @surister, @kneth

bmunkholm:
We should absolutely hesitate until it's been retested. And as you pointed out some of these were not good examples of how we would write them now. Let's please not add to the "tech dept" while we are actively trying to clean up all other pages.
It's not they they are not available to people, they are currently just in another location.

amotl:
> We should absolutely hesitate until it's been retested.

It is a natural process to fix forward. Please emphasize working on subject matters and pain points instead of working on any other pages.

- https://github.com/crate/tech-writing/issues/106

amotl:
Before closing this, let's spread some tombstone redirects, like conducted on this article.

- https://community.cratedb.com/t/autoscale-a-cratedb-cloud-cluster/1731

PR #293, opened 2025-09-14
Locust: Index page and starter tutorial

amotl:
## About

Continue adding tutorials from the community forum.

## Preview

- https://cratedb-guide--293.org.readthedocs.build/integrate/locust/

## References

- GH-102


amotl:
A user wrote in, apparently having problems with Sybase. 
```python
AttributeError: ‘QuickstartUser’ object has no attribute ‘client’
```
-- https://community.cratedb.com/t/load-testing-cratedb-using-locust/1686/2 ff.

a) Why are they using Sybase while this is about CrateDB?
b) Shall we take care of this in one way or another? /cc @hammerhead, @surister

docs/integrate/locust/tutorial.md
```diff
@@ -0,0 +1,486 @@
...
+            servers=host or CRATEDB_HOST,
+            username=CRATEDB_USERNAME,
+            password=CRATEDB_PASSWORD,
+        )
+        self._request_event = request_event
+
+    def send_query(self, sql, name, params=None):
+        cursor = self._connection.cursor()
+        start_time = time.perf_counter()
+
```
amotl, line 116: It looks like this class is defined twice in this file?

docs/integrate/locust/tutorial.md
```diff
@@ -0,0 +1,486 @@
...
```
amotl, line 314: It looks like this class is defined twice in this file?

docs/integrate/locust/tutorial.md
```diff
@@ -0,0 +1,486 @@
...
+        return response
+
+
+class CrateDBUser(User):
+    abstract = True
+
+    def __init__(self, environment):
+        super().__init__(environment)
+        self.client = CrateDBClient(self.host, request_event=environment.events.request)
+
```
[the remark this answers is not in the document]
amotl: @hammerhead: Can you confirm it is better to use `on_start` / `on_stop` with Locust? Is using `__init__` a no-go when conceiving Locust handlers?

docs/integrate/locust/tutorial.md
amotl: That's just a smaller excerpt...

docs/integrate/locust/tutorial.md
amotl: .... while this is the full version.

PR #303, opened 2025-09-15
PyCaret: Refactor section to dedicated page

amotl:
## About

Follow the lemma "each integration item on a dedicated slot within the backbone section, then link to it".

## Preview
_Rendering._


amotl:
@coderabbitai review

PR #361, opened 2025-10-03
Connect: Reorganize and clean up section

amotl:
## About
This patch sets the stage for improved guidance within the drivers section, as suggested. It is just a little reorg plus some minor copy-editing. Further improvements as discussed will be submitted using a subsequent patch.

## Preview
- https://cratedb-guide--361.org.readthedocs.build/connect/

## References
- GH-227
- https://github.com/crate/tech-content/issues/124

### DIFF
--- a/docs/connect/index.md
+++ b/docs/connect/index.md
@@ -60,6 +60,10 @@ protocol.
 - [JDBC]
 - [ODBC]
 - [SQL]
+
+```{rubric} Related
+```
+- {ref}`All drivers <connect-drivers>`
 ::::
 
 :::::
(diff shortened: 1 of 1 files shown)


bmunkholm:
In general, I'm not really sold on the concept of the links to the right at the top (and that's about all pages that has this. That could be handled separately, but might as well mention here again). The reason is my attention is drawn to them while reading the left column, not knowing what I'm missing out on and if I should investigate any of that before moving on.
So I would prefer that we in general have these references at the bottom for "more reading" after this.

I would in this case also remove "obvious" references to wikipedia and assume that's either common knowledge or that people can look it up themselves. 

Perhaps this belongs more in a general issue...

PR #362, opened 2025-10-03
Navigation: Dissolve cross-tree references

amotl:
## About
References pointing across the document tree have been reported to display confusing navigation behaviour.

## Details
Discovered per `ag "\.\.\/"`.

## Preview
- https://cratedb-guide--362.org.readthedocs.build/start/ingest.html
- https://cratedb-guide--362.org.readthedocs.build/start/query/search.html

PR #368, opened 2025-10-05
Start/Index: Refactor "Going further" to "Learn more" on index page

amotl:
## About

> "Going further" should perhaps just be moved into "First Steps" as it's basically an overview of the remaining TOC + some. Having it in the end seems very odd as that's after half the references in there.

## Preview

- https://cratedb-guide--368.org.readthedocs.build/start/

## References

- https://github.com/crate/cratedb-guide/pull/362#pullrequestreview-3302518738
- GH-227

### DIFF
--- a/docs/performance/index.md
+++ b/docs/performance/index.md
@@ -1,6 +1,6 @@
 (performance)=
 (performance-guides)=
-# Performance Guides
+# Performance guides
 
 :::{div} sd-text-muted
 Best practices and tips for sharding, scaling, and performance tuning.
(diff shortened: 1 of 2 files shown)


bmunkholm (changes requested):
I actually appreciate the speedyness! But I would keep current state than merging this as is. I have more concrete feedback, please allow me to get time for that.

amotl:
It is really just a refactoring patch that barely changes any content. Maybe other feedback beyond that would be better suited as a dedicated ticket? You can easily add your feedback to improve particular pages per "Suggest improvement" workflow.

image

-- https://cratedb.com/docs/guide/start/
-- https://cratedb.com/docs/guide/start/going-further.html

PR #382, opened 2025-10-07
README: Reflect new styles "learning path" vs. "reference material"

amotl:
## About
Improve "What's inside" conveyance in README.

## Details
Naming things: s/Guide/Handbook/

## References
- GH-372

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,23 +1,26 @@
 <div align="center">
 
-# The CrateDB Guide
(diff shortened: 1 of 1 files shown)


bmunkholm (approved):
Thanks for reflecting what we currently have.
I foresee updates to this, but it's great to reflect current state.

PR #384, opened 2025-10-07
Wording: Tame use of "This section..."

amotl:
## About

"This section [...]" was used a bit too often across the board. This patch attempts to tame it.


docs/install/container/kubernetes/kubernetes.rst
```diff
@@ -35,9 +38,8 @@ Managing Kubernetes
 
 Kubernetes deployments can be `managed`_ in many different ways. Which one
 makes sense for you will depend on your situation.
-
-This section shows you three basic commands you can use to create and update a
-resource.
+On Kubernetes, CrateDB is managed as a resource. The following commands
+will give you an idea how.
```
bmunkholm, line 42:
```suggestion
will give you an idea of how.
```

PR #424, opened 2025-10-21
Reduce marketing msg in intro page and fix link to handbook.

bmunkholm:
The marketing messaging was verbose and mostly duplicates whats already in the product pages, but at the same time don't cover as well. Let's keep docs more focused on the practical "how-to".

Related issue: https://github.com/crate/tech-content/issues/173. Still need to consider how much of the learning links should be here and how much should be moved to the Getting Started section now that that exists.

## Preview
https://cratedb-guide--424.org.readthedocs.build/


surister (approved):
lgtm

PR #453, opened 2025-10-29
Welcome: Tune down quick links

amotl:
## About
Just a bit of copy editing about layout, also tuning down card representations a bit more, for saving screen space on index pages, and improving user guidance.

## Preview
- Before: https://cratedb.com/docs/guide/
- After: https://cratedb-guide--453.org.readthedocs.build/

## Review
The patch does not include any complexities or many words to review, so the review is merely asking about your opinion if you think the page is reasonably appealing. Please share any suggestions to improve.

## References
- GH-227
- GH-446

/cc @karynzv, @hammerhead, @surister, @zolbatar, @grbade

### DIFF
--- a/docs/handbook/index.md
+++ b/docs/handbook/index.md
@@ -139,6 +139,14 @@ Use CrateDB with metrics collection agents, brokers, and stores.
 All features of CrateDB at a glance.
 ::::
 
+::::{grid-item-card}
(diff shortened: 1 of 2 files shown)


seut (approved):
thx!

bmunkholm (approved):

PR #471, opened 2025-11-03
Delete testimonials from index.md

bmunkholm:
Removed testimonials section from documentation - let's leave marketing to the webpages.

## Preview
https://app.readthedocs.org/projects/cratedb-guide/builds/30180094/

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -90,23 +90,6 @@ Use CrateDB and CrateDB Cloud in practice.
 
 :::::
 
-:::{rubric} Testimonials
-:::
-
-> When using CrateDB, a project that got started around the same time, it's like
-you've stumbled into an alternative reality where Elastic is a proper database.
->
-> <small>-– Henrik Ingo, Nyrkiö Oy, independent database consultant, MongoDB</small>
-
-> CrateDB enables use cases we couldn't satisfy with other
-database systems, also with databases which are even stronger
-focused on the time series domain.
->
-> CrateDB is not your normal database!
->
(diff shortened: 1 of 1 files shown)


bmunkholm:
Hey @amotl, I actually thought we were totally aligned about removing marketing speak from the docs, and that the purpose of the docs is to educate users on how to use and get maximum value out of the product?

You also wrote that [here](https://github.com/crate/tech-content/issues/175): _"Get rid of promotional or marketing language that is occasionally slipping in through the influx of material from other sources. Documentation should be rather dry than yappy."_.

Isn't this aligned 100% with that? Or did I misunderstand something?

I believe we agree that the Overview page should guide people to the various sections of the docs and give a general introduction to the content, agree? I think there is more to do, but wanted to start with something non-controversial :-).

Btw, for maximal credibility I don't think having a quote from a former employee is appropriate - irrespective on the timing of quote and employment. I'm quite convinced that should be removed in any case.

I also don't think these 2 quotes have any meaningful effect on our marketing results tbh. So if that's really a problem to solve in the docs, we need to align on a totally different approach and direction.

In summary, until we agree to do marketing in docs, I think we should remove this and move towards the agreed(?) direction: Docs are technical, descriptive, helpful and educate our users to get maximum value out of our product.

Thoughts?

amotl:
In general, please always feel free to make progress on the docs, by reorganizing, adding, and removing fragments for a better outcome. If one section will be removed here, it might reappear somewhere else. I mean, this repository is a work in progress and visited so many places already, so there is no need to provide any stop energy on that journey of visiting more nice spots in the future.

image

surister (approved):
LGTM

PR #477, opened 2025-11-03
Driver: Add page about Raku

amotl:
## About
What the title says.

## Preview
https://cratedb-guide--477.org.readthedocs.build/connect/raku/


bmunkholm (changes requested):
I don't think we need to should add this (and similar minority languages) with their own page.
See comments https://github.com/crate/tech-content/issues/124#issuecomment-3484870195.

PR #517, opened 2025-12-25
Maintenance: Remove `doc` as database name across the board

amotl:
## About
The default CrateDB database is called `crate`.
`doc` is certainly wrong (default schema name), so let's remove it.

## Synopsis
```diff
-postgresql://<user>:<password>@<cluster>.cratedb.net:5432/doc
+postgresql://<user>:<password>@<cluster>.cratedb.net:5432/
```

## Thoughts
If we would use `crate` instead, it would add more `crate` noise, and also suggest that the reader may use a different database name instead, which currently is not possible. So, the only viable conclusion is to remove it, unless there are different proposals.

## References
- GH-475


surister (approved):
Fair enough, nothing agains't this

PR #523, opened 2026-01-16
Unify the crate-python documentation

kneth:
## About

The idea is to move the crate-python documentation from its original repository and have it in the main documentation repository.

**This PR should not be squashed when merging**

### Disadvantages

* Adding new functionality to crate-python will require two PRs - one in the code and one in the documentation

### Advantages

* It is less likely to duplicate information in the documentation

### DIFF
--- a/docs/connect/python/crate-python.md
+++ b/docs/connect/python/crate-python.md
@@ -53,6 +53,12 @@ with conn:
     print(result)
 ```
 
+```{toctree}
+:hidden:
+
+crate-python/index
+```
+
 :::{rubric} See also
 :::
 
@@ -61,8 +67,8 @@ with conn:
 :padding: 0
 
 ::::{grid-item-card} {octicon}`book;1.75em;sd-text-info` &nbsp; Documentation
-:link: crate-python:index
-:link-type: ref
+:link: crate-python
+:link-type: url
 :link-alt: Python DBAPI driver for CrateDB
 The full documentation for the Python driver.
 ::::


joerg84:
Justtrying to understand the two separate concerns here (i.e., which problem are we solving/discussing) 
a) where to maintain the documentation (i.e., which repo)?
b) where to display the documentation (i.e., separate guide/page)?

Are those two necessarly the same, or can we pull that part from the repo when building docs (sincere question)?

bmunkholm:
I believe this is just another way to solve the problem that we currently have drivers 2 places in the TOC, as elaborated in this issue: https://github.com/crate/tech-content/issues/124. We only want 1 entry point for drivers - and I believe we all agree on that.
All options suggest that we get rid of the current "Database Drivers" section and the discussion is how to then show the docs for those drivers we removed.

There are then a few ways to do that:
1. As suggested in the footnote above, we have the docs for the individual drivers live in their own space and only linked to from the TOC. Docs are still searchable.
2. This PR suggest that we physically move the docs from the driver repos with their commits into the guide, so that the pages are preserved as they currently are (and likely the first page is adapted a bit). 
3. Alternatively (which isn't in any PR) instead of removing the docs from the repo, we could symlink the docs from the driver repos into the guide, and then be able to embed the full docs in the Connect/Drivers section.
4. We could also remove the docs form the public docs, and just have them in the repo as .md files and link to that. Docs are not searchable.

So it appears there are technical solutions to all the ways we could represent this to users the way we want to. We should probably just discuss what user experience we want for which drivers, as it appears that some drivers are so out dated that we wouldn't even recommend to use them, while others (python?) we want to promote. So we may want to do different things to different drivers?

E.g.:
1. Those we want to promote: Should have the full docs in the TOC (so they could use solution 4 above - stay in the repo, but embed in TOC with symlinks in the guide)
2. Not maintained, still useful: Just link to it with either solution 1 to make docs searchable.
3. Don't recommend: Remove entirely from docs, keep docs in .md in the repo (solution 4), explicitly to not make them searchable.

We then just need to decide which driver to treat which way. 

---

As for this PR, I guess I find the other 3 solutions better assuming sym-linking isn't a problem.

seut (changes requested):
Whatever we do to solve the TOC issues, moving the documentation out of the repository is no option at all as we do this by intent. Having the documentation inside the repository let us test the documentation using doctest which ensures that the documentation is always correct related to the current source code.

Moving the documentation out of the repository will break this, which may result in stale documentation or even worse, wrong one. Additionally, developers have to touch multiple repositories (crate-python + docu) if the source code changes and documentation adjustments are needed. I can see no real benefits but only drawbacks.

kneth:
As @amotl has [addressed](https://github.com/crate/tech-content/issues/124#issuecomment-3762533634) my concerns, this little experiment has proven not to be the way to go.

PR #586, opened 2026-05-11
Mention differences between Crate JDBC and pgJDBC

hariso:
## About

Added a note why Crate JDBC exists in the first place (even though that's more or less obvious) and also a link to the differences. It may be relevant for anyone migrating code.


hariso:
The links check failed due to:

```
(integrate/pycaret/index: line    6) broken    https://pycaret.org/ - HTTPSConnectionPool(host='pycaret.org', port=443): Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError(1, "[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Hostname mismatch, certificate is not valid for 'pycaret.org'. (_ssl.c:1081)")))
```

amotl (approved):
Thank you.

PR #598, opened 2026-07-16
Chore: Fix broken links

amotl:
A link target was removed from the tree. It is natural that we mostly discover that later. Visually, it should make no difference to remove the offending element, because it bears a `visually-hidden` style anyway.

/cc @kneth


hammerhead (approved):
Thank you!

PR #599, opened 2026-07-16
Revise rolling upgrade guide

hammerhead:
## About

We received feedback that the rolling upgrade guide was hard to follow for new users who aren't familiar yet with distributed systems. While the overall upgrade process was documented well before, there were the following issues:

* Step-by-step instructions were interrupted by lengthy educational background information on certain parameters and their implications, causing a loss of focus. Now: Focus on the flow with default settings, elaborate on possible deviations separately.
* Added an initial health check to ensure the cluster is green before starting
* Removed a warning that upgrades aren't possible from "testing versions". This is not applicable if we understand "testing versions" as the testing channel, which are regular releases. It may have targeted nightly builds, but those aren't supported anyway.
* Some of the SQL statements were outdated, querying `_node` instead of `node` from `sys.shards`, and also unnecessarily limiting their result to the schemas `doc` and `blob`.
* There was an incorrect statement about writes only succeeding if a quorum of shards are present, contradicting the documentation of [write.wait-for-active-shards](https://cratedb.com/docs/crate/reference/en/latest/sql/statements/create-table.html#write-wait-for-active-shards)
* Replaced the outdated `system` command with modern `systemctl` in examples
* The "Repeat" step was inconsistent, asking to issue `SET GLOBAL TRANSIENT "cluster.routing.allocation.enable" = 'new_primaries';` after upgrading every single node, even though it wasn't reset
* Added a post-upgrade health check to see if everything is back to green
* Provided clearer guidance when to use `min_availability` options `full`, `none`, or `primaries`

## Pending clarifications

* ~https://github.com/crate/crate/issues/19751~

## Preview

https://cratedb-guide--599.org.readthedocs.build/admin/upgrade/rolling.html

### DIFF
--- a/docs/admin/upgrade/rolling.md
+++ b/docs/admin/upgrade/rolling.md
@@ -1,118 +1,143 @@
-```{highlight} sh
-```
-
 (rolling-upgrade)=
 
 # Rolling Upgrade
 
 ## Introduction
 
-CrateDB provides an easy way to perform a rolling cluster upgrade with zero
-downtime.
+A rolling upgrade updates one CrateDB node at a time while the remaining nodes
+continue serving requests. This minimizes service disruption and avoids
+stopping the entire cluster.
 
-A rolling upgrade is possible from one version to the next feature version.
-Some examples:
+For each node, decommission it gracefully, install the new version,
+restart it, verify that it has rejoined the cluster, and then
+continue with the next node.
 
-- You can do a rolling upgrade from 4.5.x to 4.6.0
-- You can do a rolling upgrade from 4.8.x to 5.0.0
-  (because 4.8 is the last feature release within the 4. series).
-- You cannot do a rolling upgrade from x.y.z to x.(y + 3).z unless
-  the release notes explicitly mention support.
+To perform a rolling upgrade of a cluster, stop one node at a time using the
+{ref}`graceful stop <crate-reference:conf_graceful_stop>` procedure
+(see {ref}`Signal Handling <crate-reference:cli-crate-signals>`). This will allow
+CrateDB to relocate shards away from the node being decommissioned.
+Depending on the configured availability mode, replica shards may temporarily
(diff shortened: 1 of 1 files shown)


docs/admin/upgrade/rolling.md
```diff
@@ -197,90 +181,158 @@ cr> SELECT count(*) as remaining_shards from sys.shards
...
+Example for RHEL (DNF package manager):
 
-```
-$sh yum update -y crate
+```shell
+dnf update -y crate
 ```
 
 If you are in doubt how to upgrade an installed package, please refer to the
 man pages of your operating system or package manager.
```
karynzv: man page -> main page ?

hammerhead: It refers to the `man` command. Let me rephrase to simply "refer to your operating system documentation".

karynzv:
The warning mentioned below has been removed. Should we keep it? Also, maybe add a comment for nightly versions as well, something like "There's no guarantee that updating from nightly to another version, be that a new nightly or a testing/stable version works."
>Rolling upgrades are only possible if you are using a stable version of CrateDB. If you are upgrading to a testing version you must perform a full cluster restart.

hammerhead:
As mentioned in the PR's description, the term "testing versions" is unclear and the statement about rolling upgrades not working appears too general. If there is a release with particularities around upgrading, it would usually be mentioned in the corresponding release notes. That applies to both testing and stable versions. A warning to check release notes is already included.

We currently don't mention nightly builds anywhere in the documentation (only cloud touches on it). Even the tarball installation instructions point straight to regular releases. Hence my tendency to ignore it here as well to keep the guide focused.

> Removed a warning that upgrades aren't possible from "testing versions". This is not applicable if we understand "testing versions" as the testing channel, which are regular releases. It may have targeted nightly builds, but those aren't supported anyway.

karynzv (approved):

